PR #36, opened 2023-11-25
added function to validate credentials

subnet001:
New function will allow users to validate their credentials and return a bool depending on validity

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,6 +208,16 @@ credentials = Credentials()
 download_features(features, "/some/download/path", {"credentials": credentials})
 ```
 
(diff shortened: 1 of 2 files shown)


pedep:
Very useful. I would love to see a snippet added to the `README.md` on how to use it. If you're feeling up to it, you could also add support for it in the `cli.py`. If not, that will probably have to be added to the roadmap

pedep (changes requested):
Could you add the readme changes from above before i merge?

pedep (approved):

PR #53, opened 2023-12-15
dependabot: use for Github actions

pjonsson:
Add the github-actions package-ecosystem
to dependabot so CI actions are kept
up to date.

Actions aren't typically updated multiple
times per week, so run the check on a daily
basis.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,3 +9,7 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"


pedep:
Thank you very much for your contribution :)

I have a quick question regarding your description
> Actions aren't typically updated multiple
times per week, so run the check on a daily
basis.

If actions are not updated more than once per week, would it not make sense to run the check once per week?
I don't mind either way, as im not paying for the CI minutes anyways :wink: 

If you still think its fine as-is, then i will merge it :+1:

pjonsson:
> If actions are not updated more than once per week, would it not make sense to run the check once per week? I don't mind either way, as im not paying for the CI minutes anyways 😉

It wasn't a well-formulated description, but the point was that this change will not cause dependabot to spam your PR queue with pull requests.

> If you still think its fine as-is, then i will merge it 👍

As you say, you're not charged per minute for the CI, so I think this can be merged.

pedep (approved):
LGTM :+1:

PR #73, opened 2024-02-01
Make geojson_to_wkt function handle single-feature FeatureCollections.

aslanispavlos:
Minor change for geojson_to_wkt.

The motivation for this change is to make geojson_to_wkt work with single-feature FeatureCollections geojson files. A quite common way of creating geojson files of AOI's is by using the geojson.io webpage, which defaults to FeatureCollection even for one feature.

### DIFF
--- a/src/cdsetool/query.py
+++ b/src/cdsetool/query.py
@@ -114,6 +114,8 @@ def geojson_to_wkt(geojson):
 
     if geojson.get("type") == "Feature":
         geojson = geojson["geometry"]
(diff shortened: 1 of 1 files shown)


pedep:
Thank you very much for your contribution :1st_place_medal: 

Before i merge, could i get you to make a quick unit test for this feature? Simply copying and modifying the test in https://github.com/SDFIdk/CDSETool/blob/main/tests/query/query_test.py#L103 would be plenty

pedep (approved):
LGTM :+1:

This will be released shortly as a part of v.0.2.4 :rocket:

PR #82, opened 2024-02-25
credentials: set expire based on post time

pjonsson:
Get the current time right after the post
request returns, and calculate the expire
times from that point in time.

This minimizes the error of the token expire
times.


pjonsson:
I don't know how the authentication works, is it safe to put the `now = ` line before the post to get a pessimistic expiry time?

pedep:
> I don't know how the authentication works, is it safe to put the `now = ` line before the post to get a pessimistic expiry time?

That should be fine. 

If we want to be more precise, the `iat`, `nbf` and `exp` claims of the token should be used instead

pedep (approved):

PR #83, opened 2024-02-25
Add optional logging

pjonsson:
This adds support for clients to pass in
a configured logger object to get warning
logs on retries, and debug logs in general
when downloading.

The default logger is a no-op logger, so
default behavior remains unchanged.

### DIFF
--- a/src/cdsetool/download.py
+++ b/src/cdsetool/download.py
@@ -12,6 +12,7 @@ import time
 import shutil
 from cdsetool._processing import _concurrent_process
 from cdsetool.credentials import Credentials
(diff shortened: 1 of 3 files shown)


pjonsson:
I don't have a container set up for CDSETool so I have not tested the regular cli or the NoopLogger, I would appreciate some help with testing that.

The library path with a logger passed in to `download_feature` has been tested by downloading ~1500 products without issues.

pedep:
Thank you for yet another great contribution.
I will check out your branch and add some tests for NoopLogger :+1:

pedep (approved):

Issue #67, opened 2024-01-24
download_features returns many incomplete .zip files


massimozanetti:
Hi. and thank you for develping CDSEtool, it is something really important.

I have tried it to download some Sentinel-2 products, and after a couple of runs I discovered that many products are not complete after download, and this happens randomly. See pictures two different instances of downloading the same list of products:

run 1
![brokenzip](https://github.com/SDFIdk/CDSETool/assets/129863715/c73e94d9-ee46-448e-866f-2557405d8535)

run 2
![brokenzip2](https://github.com/SDFIdk/CDSETool/assets/129863715/06757256-fc53-4b8e-83e7-859067e46ae0)

Many of the m are clearly incomplete, and they are not the same in the two runs...
Thanks

pedep:
Hi @massimozanetti,

Thank you for your bug report
This is definitely a critical problem that should be adressed immediately :skull: 

I will try to reproduce your issue to see if something can be done. 
If i am lucky, there is some metadata in the headers from the download server, which will allow us to detect incomplete downloads. If not, i might have to quickly implement #40 in hopes of solving the issue.

In the meantime, could i get you to share a bit about the environment you are running `cdsetool` in?
- Operating system & version
- Python version, package manager and environment (conda, pip, etc)
- cdsetool version (`pip show cdsetool` or similar)
- Which internet connection is used? (corporate, domestic, university, LTE, etc)
  - Is the internet connection wired or wireless?
- Do you frequently experience network issues or package loss?
- If possible, could you share the script you are running?
  - If you cannot share your script, could you create a minimal script that recreates the issue?

Thank you for your assistance

massimozanetti:
Hi @pedep, I am running on:

- Ubuntu 18.04.6 LTS
- conda 4.14.0
- pip install cdsetool==0.2.3
- Wired Eduroam internet (institution: https://www.fbk.eu/en/)

I do not experience data loss, I strongly believe the problem is connected to the procedure documented by CDSE. The chuncked download with no checks returns this kind of problems if naively implemented as suggested here: https://documentation.dataspace.copernicus.eu/APIs/OData.html

To reproduce the problem it just takes reproducing your documented snippets, with the list of products I am interested in as input, here it is for your convenience:
```
from cdsetool.credentials import Credentials
from cdsetool.query import query_features
from cdsetool.download import download_features

username = ..........
password = .........
products_dir = .........

[...]

PR #87, opened 2024-02-26
download: use existing open fd

pjonsson:
Use the existing open file descriptor
instead of opening a new file descriptor
to the same file.

It is not clear to me if this has any
bearing for #67, but having fewer file
descriptors open will not make anything
worse.


pedep:
@pjonsson Thank you :star2:

pedep (approved):

PR #90, opened 2024-03-05
download: write all received bytes to file

pjonsson:
This is what `stream_response_to_file` from the requests toolbelt package does, and that package resides alongside the requests package on Github.

Refs #67


pedep:
Thank you for researching and implementing a solution for #67 :rocket:

pedep (approved):

PR #92, opened 2024-03-06
proxy feature on query and download

LattaruoloAndrea:
Hello, I have made the changes just for the proxy feature, it can be used it the following way:

```Python
proxies = {'http': 'http_proxy', 'https': 'https_proxy'}

features = query_features(
    "Sentinel1",
    {
        "startDate": "2020-12-20",
        "completionDate": date(2020, 12, 25),
        "processingLevel": "LEVEL1",
        "sensorMode": "IW",
        "productType": "IW_GRDH_1S",
        "geometry": shape_to_wkt("path/to/shapefile.shp"),
    },
    proxies
)

list(
    download_features(
        features,
        "path/to/output/folder/",
        {
            "concurrency": 4,
            "monitor": StatusMonitor(),
            "credentials": Credentials("username", "password"),
            "proxies": proxies,
        },
    )
)
```

I tried to add the proxy to all the requests in the query and added the proxy in the session object in the download part, let me know if it work this time.
if the fields of proxies are not passed in the query and download it should work as it was before.

### DIFF
--- a/src/cdsetool/query.py
+++ b/src/cdsetool/query.py
@@ -42,10 +42,12 @@ class FeatureQuery:
 
     total_results = None
 
-    def __init__(self, collection, search_terms, proxies = {}):
+    def __init__(self, collection, search_terms, proxies={}):
         self.features = []
         self.proxies = proxies
-        self.next_url = _query_url(collection, {**search_terms, "exactCount": "1"}, proxies)
+        self.next_url = _query_url(
+            collection, {**search_terms, "exactCount": "1"}, proxies
+        )
 
     def __iter__(self):
         return _FeatureIterator(self)
@@ -228,14 +230,15 @@ def _assert_max_inclusive(search_term, max_inclusive):
 _describe_docs = {}
 
 
-def _get_describe_doc(collection, proxies = {}):
+def _get_describe_doc(collection, proxies={}):
     if _describe_docs.get(collection):
         return _describe_docs.get(collection)
 
     res = requests.get(
         "https://catalogue.dataspace.copernicus.eu"
         + f"/resto/api/collections/{collection}/describe.xml",
-        timeout=120, proxies=proxies
+        timeout=120,
+        proxies=proxies,
     )
     assert res.status_code == 200, (
         f"Unable to find collection with name {collection}. Please see "
--- a/src/cdsetool/query.py
+++ b/src/cdsetool/query.py
@@ -42,10 +42,10 @@ class FeatureQuery:
 
     total_results = None
 
-    def __init__(self, collection, search_terms,proxies = {}):
+    def __init__(self, collection, search_terms, proxies = {}):
         self.features = []
         self.proxies = proxies
-        self.next_url = _query_url(collection, {**search_terms, "exactCount": "1"},proxies)
+        self.next_url = _query_url(collection, {**search_terms, "exactCount": "1"}, proxies)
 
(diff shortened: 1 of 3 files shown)


src/cdsetool/query.py
```diff
@@ -42,9 +42,10 @@ class FeatureQuery:
 
     total_results = None
 
-    def __init__(self, collection, search_terms):
+    def __init__(self, collection, search_terms,proxies = {}):
```
pjonsson, line 45: Try to run `black src` from the top directory again, I'm pretty sure there's a missing space between `,` and `proxies`.

LattaruoloAndrea: added all the spaces after the , in the proxy that I added

pedep:
@LattaruoloAndrea pylint has found a few issues with your code. If you fix those, i will merge and release immediately :nerd_face: :+1:

LattaruoloAndrea:
> @LattaruoloAndrea pylint has found a few issues with your code. If you fix those, i will merge and release immediately 🤓 👍

I should have fixed.

pedep (approved):

pedep:
After having spent a few minutes fixing the last pylint issues, i noticed that `query_features` does not pass the `proxies` argument to the `FeatureQuery` class. This should probably be adressed.

@subnet001 @pjonsson Having looked long at the spaghetti we are weaving trying to pass arguments from the api function to the util functions that actually need them, i think we should have a look at separating some of this logic into a purpose-built class.

I was thinking something along the lines of this:

```python

# http.py
class HTTP
  proxies = None

  def get(...):
    requests.get(..., proxies = HTTP.proxies)
    
# main.py

from cdsetool.http import HTTP

HTTP.proxies = {
  # proxy conf
}

query_features("Sentinel2") # uses HTTP as backend instead of requests
```

This way we pass the configs directly into where they are needed via the static variable, instead of having to pass along variables through half a dozen methods. Only downside is no per-request proxies.

pjonsson:
@pedep without commenting on the technical parts of your suggestion, I think that falls outside the scope of this PR, so perhaps a ticket should be filed for that?

I can't immediately see the ramifications of the suggestion for me, so let me ponder them and get back to you on that.

pedep:
> I think that falls outside the scope of this PR, so perhaps a ticket should be filed for that?

Very true, i will move this to a discussion or issue.

> I can't immediately see the ramifications of the suggestion for me, so let me ponder them and get back to you on that.

I will link the issue here for continuity :+1: Your input is appreciated 

@LattaruoloAndrea To finish up your pull request, i think we just need a way for `query_features` in `query.py` to accept `proxies` so that it may be passed on to the `FeatureQuery`

LattaruoloAndrea:
> @LattaruoloAndrea To finish up your pull request, i think we just need a way for `query_features` in `query.py` to accept `proxies` so that it may be passed on to the `FeatureQuery`
That's true I actually missed that 
```Python
def query_features(collection, search_terms, proxies=None):
    """
    Returns an iterator over the features matching the search terms
    """
    return FeatureQuery(collection, {"maxRecords": 2000, **search_terms}, proxies)
```
This change should do the trick. 

> I was thinking something along the lines of this:

This is a great way, personally when I implemented the proxy I tried to not introduce many code, but I agree the "pass through" from funtion to funtion is not ideal.

pedep (approved):

Issue #69, opened 2024-01-26
Restart a partial download


LattaruoloAndrea:
## Summary
Hello, since the data that we download it can be demanding it would be a great idea to allow restarting partial dowloads. 

### Basic example 
Something like the code below could work 
For example a file it's been stopped with M bytes downloaded, we could do a request on dowload starting from M+1 bytes
```Python
        already_downloaded_bytes = 0
        headers = {"Authorization": f"Bearer {access_token}"}
        if os.exists(relative_path):
            already_downloaded_bytes =relative_path.stat().st_size # Get the M bytes already donwloaded
            headers = {"Authorization": f"Bearer {access_token}","Range": "bytes={}-".format(already_downloaded_bytes)}
``` 
The only issue that I have found is that every time I tried to use the header with bytes that was different from 0 the api gave me and error. I don't know if it's possible to achive partial downloads but it would be useful. 

###Where to save the partial files
I have seen that the partial files during the dowload are created with 
```Python
fd, tmp = tempfile.mkstemp()
```
I guess it would be better to create a partial file in the path that we insert adding a '.incomplete' to the name of the file 
```Python
def download_feature(feature, path, options=None):
    """
    Download a single feature

    Returns the feature ID
    """
    options = options or {}
    url = _get_feature_url(feature)
    filename = feature.get("properties").get("title")

    if not url or not filename:
        return feature.get("id")

    overwrite = options.get("overwrite_existing", False)
    continue_ = options.get("continue_existing", True)

    if not os.path.exists(Path(path)):
        os.makedirs(path)
    result_path = os.path.join(path, filename.replace(".SAFE", ".zip"))
    temp_path =  Path(path) / filename.replace(".SAFE", ".zip.incomplete") #os.path.join(path, filename.replace(".SAFE", ".zip.incomplete"))
    if not overwrite and os.path.exists(result_path):
        return feature.get("id")
    
    if not continue_ and os.path.exists(temp_path):
        #TODO decide what to do with partlial files, 
        """
        it should be appropriate to restart the download from the
        from where it let off istead of the beginning.
        The current version remove the partial file and restart the download
        from the beginning.
        """
        return feature.get("id")
    if os.path.exists(temp_path):
        os.remove(temp_path)
```
like this is the implementation that I made of where to dave the partial downloads on the download_feature() function

pedep:
Hi @LattaruoloAndrea,

Thank you for your feature request :1st_place_medal:

[...]

Issue #40, opened 2023-11-26
Checksum validation of downloaded files


pedep:
## Summary
If network issues occur, the downloaded files could be corrupted. This can easily be checked by using a checksum

### Basic example
The OpenSearch API does not include checksums, but luckily the OData API includes the checksum, and can be searched for invididual features
https://catalogue.dataspace.copernicus.eu/odata/v1/Products?$filter=Id%20eq%20%27b30f2bbf-7f02-494a-b3f4-3ea11cd3d14d%27

### Motivation
To guard against corrupt files on download.
It should be possible to hash chunks sequentially as they are downloaded, to avoid extra disk IO

shuoO-24:
Hi @pedep , I would like to work on this issue. Can you please assign it to me and elaborate more on how to get the checksum from the OData API? Thanks very much!

massimozanetti:
Hi all, regarding this, how this feature handles the fact that in CDSE many (actually the most) products have "null" value for `checksum` and `contentlength` fields?

pedep:
@massimozanetti Do you have an example of a product with no `checksum` and `contentlength`?

massimozanetti:
> @massimozanetti Do you have an example of a product with no `checksum` and `contentlength`?

Hi @pedep, sure I can show you many examples. Here are examples (with little python code to plot the results):

`import requests
import pandas as pd

response = requests.get("https://catalogue.dataspace.copernicus.eu/odata/v1/Products?$filter=contains(Name,'S1A_IW_SLC') and ContentDate/Start gt 2022-05-03T00:00:00.000Z and ContentDate/Start lt 2022-05-21T00:00:00.000Z&$expand=Attributes") #len=0, chk=[]

#response = requests.get("https://catalogue.dataspace.copernicus.eu/odata/v1/Products?$filter=contains(Name,'S2A_MSI') and ContentDate/Start gt 2022-05-03T00:00:00.000Z and ContentDate/Start lt 2022-05-21T00:00:00.000Z&$expand=Attributes") #len=0, chk=[]

#response = requests.get("https://catalogue.dataspace.copernicus.eu/odata/v1/Products?$filter=contains(Name,'S2A_MSI') and ContentDate/Start gt 2023-05-03T00:00:00.000Z and ContentDate/Start lt 2023-05-21T00:00:00.000Z&$expand=Attributes") #len=wrong value, chk=[{}]

json = response.json()

df = pd.DataFrame.from_dict(json['value'])


[...]

PR #96, opened 2024-03-08
features checksum

LattaruoloAndrea:
Hello I tried to implement the checksum.
In ordet to do the validity for the checksum I added in the FeaturesQuery.features three fields that are taken from the odata API,
these fileds are:
- Checksum (the data might not be present as far as I know as discussd in `[#40)](https://github.com/SDFIdk/CDSETool/issues/40#issuecomment-1931868462) 
```Python
        for product_odata in odata_list_value:
            map_id_checksum[product_odata["Id"]] = {
                "Checksum": product_odata.get("Checksum", []), #here if present takes it otherwise empty list 
                ...
            }
```
- ContentLength (I do a check if contentLenght is the same as the size in the features.propertie.services.download.size but it might be redundant)
```Python
                if (
                    feature.get("properties")
                    .get("services")
                    .get("download")
                    .get("size")
                    == product_odata["ContentLength"]
                ):
                    feature["ContentLength"] = product_odata["ContentLength"]
                else:
                    print(
                        f"Warning: content length of {feature.get('id')} does not match the ContentLength in odata"
                    )
```
- odata (the whole odata not sure if it could be useful to someone)

Then on the download part I left a few TODO since I am not sure of the logic to implement in these cases. It can be discussed.

First question:
```Python
        valid_checksum = validity_check(tmp, feature)
        if valid_checksum == Validity.VALID or valid_checksum == Validity.IGNORE:
            shutil.move(tmp, result_path)
        else:
            # TODO here to implement the logic what if the checksum is invalid?
            os.remove(tmp)
```
after the validity check, which can be otputed in 4 different state:

```Python
class Validity(Enum):
    VALID = 1 # the file is not corrupted
    INVALID = 2 # the file is corrupted 
    IGNORE = 3 # verification is ignored for lack of metadata
    CONTINUE = 4 # it's a partial file (so far it's handled as a corrupted one)
```

what I have done so far is if the file is valid or metadata is not present we move the temp in the result otherwise we delete the dowloaded stuff (the problem here is there is no feedback of what is not downloaded), 

As I seggested on a previous pull request it could be usefull to integrate a sort of retries attemp in the logic of download_feature() as mentioned in [#72](https://github.com/SDFIdk/CDSETool/pull/72#discussion_r1482612361) so we can handle corrupted files and partial ones as discussed in [#69](https://github.com/SDFIdk/CDSETool/issues/69). So far I tried to stick just on the checksum and integrity of the file.


src/cdsetool/download.py
```diff
@@ -126,3 +143,60 @@ def _set_proxy(options, session):
...
+def validity_check(temp_path, product_info):
+    """Given it's temp_path and metadata info checks if the data downloaded is valid"""
+    size = os.path.getsize(temp_path)
+    content_length = (
+        product_info.get("properties").get("services").get("download").get("size")
+    )
+    if size > content_length:
+        # Dowloaded more than the size of the file.
+        return Validity.INVALID
+    if size == content_length:
```
pjonsson, line 157: If you change the condition to `if size != content_length: return Validity.CONTINUE` you can save a nesting level for the following code.

src/cdsetool/download.py
```diff
@@ -126,3 +143,60 @@ def _set_proxy(options, session):
...
+    if checksum is None:
+        # no checksum available
+        return None
+    with open(temp_path, "rb") as f:
+        while True:
+            block_data = f.read(8192)
+            if not block_data:
+                break
+            algo.update(block_data)
+    return algo.hexdigest().lower() == checksum.lower()
```
pjonsson, line 185: Should the `.lower()` really be there?

I think changing line 178 to `return Validity.IGNORE` and this line to `return Validity.VALID if algo.hexdigest().lower() == checksum.lower() else Validity.INVALID` would eliminate the need for the logic on lines 159-165.

src/cdsetool/download.py
```diff
@@ -55,9 +69,12 @@ def download_feature(feature, path, options=None):
             for chunk in response.iter_content(chunk_size=1024 * 1024 * 5):
                 file.write(chunk)
                 status.add_progress(len(chunk))
-
-        shutil.move(tmp, result_path)
-
+        valid_checksum = validity_check(tmp, feature)
+        if valid_checksum in (Validity.VALID, Validity.IGNORE):
```
pjonsson, line 73: I have no idea if `not in` works, but you can save an intermediate variable and indentation by inverting the condition:

```python
if validity_check(tmp, feature) not in (Validity.VALID, Validity.IGNORE):
    log.error(f"Faulty checksum for {filename}")
    os.remove(tmp)
    return None
```

pyproject.toml
```diff
@@ -21,6 +21,7 @@ dependencies = [
   "requests >= 2.28.1,< 2.32.0",
   "pyjwt[crypto] ~= 2.8.0",
   "geopandas >= 0.13.2",
+  "blake3 >=0.4.1",
```
pjonsson, line 24: The other dependencies have a space between `>=` and the version number.

src/cdsetool/download.py
LattaruoloAndrea: I guess your suggesting would point out that we don't want to consider to restart downloading partial files (the `Validity.Continue` I inteded to use it just in case of restarting partial files and not for corrupted ones)?
for example we download the first 5000 Bytes then by checking the validiy I used `Validity.Continue` for retry the download, but this part is still am Idea.

src/cdsetool/query.py
```diff
@@ -84,6 +86,63 @@ def __set_next_url(self, res):
...
+                if (
+                    feature.get("properties")
+                    .get("services")
+                    .get("download")
+                    .get("size")
+                    == product_odata["ContentLength"]
+                ):
+                    feature["ContentLength"] = product_odata["ContentLength"]
+                else:
+                    print(f"Warning: {feature.get('id')} no match in sizes")
```
pjonsson, line 142: Should `feature["odata"]` and `feature["Checksum"]` be set if the sizes differ?

src/cdsetool/query.py
```diff
@@ -84,6 +86,63 @@ def __set_next_url(self, res):
...
+            else:
+                print(f"Warning: {feature.get('id')} not in odata")
```
pjonsson, line 144: There's only a no-op logger available in the tree, but it would be nice if these warnings were calling the logger in a similar way that is done by `download_feature`.

src/cdsetool/query.py
```diff
@@ -72,7 +72,9 @@ def __fetch_features(self):
             total_results = res.get("properties", {}).get("totalResults")
             if total_results is not None:
                 self.total_results = total_results
-
+            odata = self.__fetch_odata_from_product_list(self.features)
```
pjonsson, line 75: The `odata` value is never used outside of `__add_odata_to_features()`, so call  `self.__add_odata_to_features(self.__fetch_odata_from_product_list(self.fetures))` here and put the if-statement first in `__add_odata_to_features()` instead.

src/cdsetool/download.py
pjonsson: My suggestion doesn't change the logic as far as I can tell:
```python
if size != content_length:
    return Validity.CONTINUE
checksum_comparison = _checksum_compare(temp_path, product_info)
if checksum_comparison is None:
  # No data available for the checksum
  return Validity.IGNORE
if checksum_comparison:
  return Validity.VALID
# Checksum failed
return Validity.INVALID
```

src/cdsetool/download.py
```diff
@@ -126,3 +143,60 @@ def _set_proxy(options, session):
...
+            return Validity.VALID
+        # Checksum failed
+        return Validity.INVALID
+    # here it's a partial download logic to implement is TODO
+    return Validity.CONTINUE
+
+
+def _checksum_compare(temp_path, product_info):
+    """Compare a given MD5 checksum with one calculated from a file."""
+    checksum = None
```
pjonsson, line 172: Remove this and the next line, `algo` and `checksum` are both initialized on line 175.

src/cdsetool/download.py
LattaruoloAndrea: The problem with using 
```Python
if size != content_length:
    return Validity.CONTINUE
``` 
is that you enter that condition two times if you donloaded more of the actual size or less, which are two different logics. If you downloaded more than the file is corrupted if you donwloaded less then you can resume the donwload (but this logic is yet to implement or if we want to implement as suggested in [Restart a partial download)](https://github.com/SDFIdk/CDSETool/issues/69)

```Python 
    if size > content_length:
        return Validity.INVALID
    if size == content_length:
        return _checksum_compare(temp_path, product_info)
    # here it's a partial download logic to implement is TODO
    return Validity.CONTINUE
```

current implementation as suggested in 

> I think changing line 178 to `return Validity.IGNORE` and this line to `return Validity.VALID if algo.hexdigest().lower() == checksum.lower() else Validity.INVALID` would eliminate the need for the logic on lines 159-165.

src/cdsetool/query.py
pjonsson: I don't have an answer. If there isn't a clear right or wrong, there should probably be a comment in the code describing why the implemented behavior was chosen.

src/cdsetool/download.py
```diff
@@ -55,9 +69,10 @@ def download_feature(feature, path, options=None):
             for chunk in response.iter_content(chunk_size=1024 * 1024 * 5):
                 file.write(chunk)
                 status.add_progress(len(chunk))
-
+        if validity_check(tmp, feature) not in (Validity.VALID, Validity.IGNORE):
+            log.error(f"Faulty checksum for {filename}")
+            os.remove(tmp)
```
pjonsson, line 74: Missing `return None` after this line.

src/cdsetool/download.py
pjonsson: Sorry, I don't understand what you are saying. Which part of this doesn't work?
```python
if size > content_length:
  return Validity.INVALID
if size != content_length:
  # here it's a partial download logic to implement is TODO
  return Validity.CONTINUE
return _checksum_compare(temp_path, product_info)
```
And seeing that code spelled out, the condition should probably be `<` rather than `!=`.

src/cdsetool/query.py
LattaruoloAndrea: Sounds good to me I changed the implementation to: 
```Python
                    product_odata = map_id_checksum[feature.get("id")]
                    # we take the odata and checksums even if they are wrong 
                    # then they are handled by the downloader
                    feature["odata"] = product_odata["odata"]
                    feature["Checksum"] = product_odata["Checksum"]
                    if (
                        feature.get("properties")
                        .get("services")
                        .get("download")
                        .get("size")
                        != product_odata["ContentLength"]
                    ):
                        # There is problems with the metadata provided by esa as far as
                        # we know in https://github.com/SDFIdk/CDSETool/issues/40#issuecomment-1931868462
                        # the sizes could differ but we will still add the odata and checksum to the features
                        # if the Checksum provided is wrong it's handled by the downloader
                        self.logger.warning(
                            f"Warning: {feature.get('id')} no match in sizes"
                        )
```
I added the comments explaining the logic and the differ in the sizes it's logged as a worning

src/cdsetool/download.py
LattaruoloAndrea: AH ok I see I was missing some parts before this one works for me I will add now

LattaruoloAndrea:
@pjonsson Hi how do you run the test in local with pytest? So that I can impement the mock for the Odata so there is less test failures

pjonsson:
I have no idea, but `.github/workflows/pytest.yml ` uses `pytest --cov=cdsetool --cov-report xml`. I'm guessing you can omit the cov-parameters if you are not interested in coverage information.

LattaruoloAndrea:
Hello @pjonsson, @pedep  I guess I was able to solve all the problems on this feature, maybe it's better to check wheter I did not change any logic in [96](https://github.com/SDFIdk/CDSETool/pull/96/commits/9b43f878ea3e866ffbca5682840eb93d32f24d41)

PR #97, opened 2024-03-10
credentials: inline private methods

pjonsson:
Initialize the expire times for credentials to a time, and inline a few single-use methods that are private to the class. This saves almost 50 lines of code and reduces the surface area of the Credentials object while preserving existing functionality.

### DIFF
--- a/src/cdsetool/credentials.py
+++ b/src/cdsetool/credentials.py
@@ -93,25 +93,6 @@ class Credentials:  # pylint: disable=too-few-public-methods disable=too-many-in
         session.headers.update({"Authorization": f"Bearer {self.__access_token}"})
         return session
 
-    def __exchange_credentials(self):
-        data = {
-            "grant_type": "password",
-            "username": self.__username,
-            "password": self.__password,
-            "client_id": "cdse-public",
-        }
-
-        self.__token_exchange(data)
-
-    def __refresh_access_token(self):
-        data = {
-            "grant_type": "refresh_token",
-            "refresh_token": self.__refresh_token,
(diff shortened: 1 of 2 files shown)


src/cdsetool/credentials.py
```diff
@@ -22,12 +22,6 @@ class InvalidCredentialsException(Exception):
     """
 
 
-class NoTokenException(Exception):
-    """
-    Raised when no token is available
-    """
-
-
```
pedep, line 30: Could you do something like this, to make for a safer release?

```python

class DeprecatedNoTokenException(Exception):
    """
    deprecated
    """
def NoTokenException(*args, **kwargs):
    from warnings import warn
    warn("Warning! NoTokenException is deprecated, and will be removed in the next major release.")
    return DeprecatedNoTokenException(*args, **kwargs)
```

pedep (changes requested):
LGTM besides my request above :+1:

src/cdsetool/credentials.py
pjonsson: It seems fair to remain somewhat compatible for a while.

I have no idea how your code works under the hood, but I added the lines with a capitalized "D" in the pydoc.

pedep: Its essentially just a wrapper around the constructor of the old class, which outputs the warning message before running the old class constructor.
Just to help users transition, instead of receiving import errors in their code.

pjonsson: Ok, fixed, and passed linting now too.

pedep (approved):

Issue #93, opened 2024-03-07
HTTP wrapper class w. global configuration


pedep:
After having spent a few minutes fixing the last pylint issues, i noticed that `query_features` does not pass the `proxies` argument to the `FeatureQuery` class. This should probably be adressed.

@subnet001 @pjonsson Having looked long at the spaghetti we are weaving trying to pass arguments from the api function to the util functions that actually need them, i think we should consider separating some of this logic into a purpose-built class.

I was thinking something along the lines of this:

```python

# http.py
class HTTP
  proxies = None

  def get(...):
    requests.get(..., proxies = HTTP.proxies)
    
# main.py

from cdsetool.http import HTTP

HTTP.proxies = {
  # proxy conf
}

query_features("Sentinel2") # uses HTTP as backend instead of requests
```

This way we pass the configs directly into where they are needed via the static variable, instead of having to pass along variables through half a dozen methods. Only downside is no per-request proxies.

_Originally posted by @pedep in https://github.com/SDFIdk/CDSETool/issues/92#issuecomment-1983307927_

pjonsson:
I had a look at the code and there are a few things I don't understand:
1. When `Credentials` is constructed, `__exchange_credentials()` is called and that performs HTTP requests. Since that happens first in `download_features`, those requests are made without a HTTP proxy, which is likely to go straight into the company firewall and fail. So `Credentials` constructor should take a proxies parameter as well? If it did, `get_session()` could configure the proxy before returning the session, and the proxy code from download.py could be removed.
1. In query.py, `_query_url()` has a default value for proxies. Since it is an internal function, and the only in-tree caller already has a default value for its own proxies parameter, nothing will ever use the default value of `_query_url`, and same comment for `_get_describe_doc`.

I can't see why someone would need to have different proxies for each request, my impression is that if you are stuck behind a proxy, you need it for all requests.

I just started learning Python so I'm not sure I understand the details of the suggestion in this issue, but `HTTP` is never passed to `query_features`, so it must be some kind of global variable? Personally, I'm not stuck behind a proxy, so the solution for proxies can be arbitrarily bad without making me unhappy, but my experience is that with time, more things will end up inside the HTTP object since it is already there, and all that state will make things harder to test, which in turn will become a maintenance issue.

[...]

PR #99, opened 2024-03-11
credentials: retry calls on failures

pjonsson:
Add a retry mechanism for the session
created in get_session() and use that
inside the Credentials object.

The Retry class from urllib3 respects
the Retry-After header, so that should
hopefully avoid these types of failures
in the future:

HTTPError: 503 Server Error: Service Unavailable for url: https://identity.dataspace.copernicus.eu/auth/realms/CDSE/.well-known/openid-configuration

which I saw raised inside __openid_configuration().

### DIFF
--- a/src/cdsetool/credentials.py
+++ b/src/cdsetool/credentials.py
@@ -8,6 +8,8 @@ import netrc
 import threading
 import requests
 import jwt
+from requests.adapters import HTTPAdapter
+from urllib3.util import Retry
 
 
 class NoCredentialsException(Exception):
@@ -63,6 +65,12 @@ class Credentials:  # pylint: disable=too-few-public-methods disable=too-many-in
         self.__username = username
         self.__password = password
 
+        self.__retries = Retry(
+            total=5,
+            backoff_factor=0.5,
+            raise_on_status=False,
+            status_forcelist=Retry.RETRY_AFTER_STATUS_CODES,
+        )
         self.__openid_conf = None
         self.__jwks = None
         self.__openid_configuration_endpoint = (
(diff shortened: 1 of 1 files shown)


pjonsson:
This had some light testing (224 downloaded products) without anything obvious failing. Worst case is that I did something wrong with the configuration and this does nothing more than what is already done today.

src/cdsetool/credentials.py
```diff
@@ -83,19 +91,37 @@ def __init__(
...
+        session = self.get_session(authorization=False)
+        retries = Retry(
+            total=2,
+            backoff_factor=0.5,
+            allowed_methods=None,
+            raise_on_status=False,
+            status_forcelist=Retry.RETRY_AFTER_STATUS_CODES,
+        )
+        session.mount("http://", HTTPAdapter(max_retries=retries))
+        session.mount("https://", HTTPAdapter(max_retries=retries))
```
pedep, line 121: I'm not sure i see the point in modifying the `get_session` method to be usable here (the `authorization=True` addition on L94) when you end up repeating the bulk of the `get_session` method here. 

This could be slimmed down with no difference in function by reverting the changes made to `get_session` and applying this change
```suggestion
        session = requests.Session()
        retries = Retry(
            total=2,
            backoff_factor=0.5,
            allowed_methods=None,
            raise_on_status=False,
            status_forcelist=Retry.RETRY_AFTER_STATUS_CODES,
        )
        session.mount("http://", HTTPAdapter(max_retries=retries))
        session.mount("https://", HTTPAdapter(max_retries=retries))
```

pedep (changes requested):
Would there be any harm in using the same retry config internally in `Credentials` as well as the session returned by `get_session`?

src/cdsetool/credentials.py
pjonsson: You have a point about code duplication, but line 183 uses `get_session`, and so does `download_feature`, so I don't think those changes should be reverted.

How about passing in an optional Retry object to get_session instead? That way the code you marked can retry a little and for any method, and the rest of CDSETool uses the (presumably) safe defaults from urllib3?

Pushed code is completely untested, I have no idea if I can assign parameter values with `retries=retries or self.__retries`.

pedep: >  but line 183 uses get_session, and so does download_feature, so I don't think those changes should be reverted.

I can see where you are going with this, but i think the reusable session code should be moved elsewhere. The entire point of the `Credentials` class is to generate sessions with authorization enabled, and the `get_session` method has this implication built in.
Adding the option to disable this more or less bypasses the class's entire function.

I think the `get_session` code as it is right now would fit better in a generic http backend class along the lines of #93 . This also has the additional benefit of providing a convenient way of configuring retries globally, through a mechanism similar to the suggested global `proxies` config. 

I would either go the way of the HTTP class, or provide a private method for generating sessions with retry and using that in the public `get_session` method. This way the public API of the `Credentials` class does not get tainted by private code.

pjonsson: I think the HTTP class is a much bigger change, so I have updated this PR to keep the internal use in a private method as you suggest.

pedep (approved):

Issue #84, opened 2024-02-25
Hanging download_feature because of 401/404 response


pjonsson:
## Describe the bug
After some time downloading, the download hangs inside `download_feature`. I am not sure if the hang is deterministic, but it happens several times per week.

With #83 on top of main, here is a partial log with correct timestamps:
```
time=2024-02-25T04:59:48 level=INFO msg="Download started"
<614 successful downloads>
time=2024-02-25T13:02:48 level=INFO msg="Download started"
time=2024-02-25T13:02:49 level=WARNING msg="Status code 401, retrying.." file=download.py:109
time=2024-02-25T13:03:58 level=WARNING msg="Status code 401, retrying.." file=download.py:109
<loop continues with 401 status>
```
So it's hanging in the loop at https://github.com/SDFIdk/CDSETool/blob/8385e3a82d0915f472458955608ec40aadb0e082/src/cdsetool/download.py#L103

This is a different issue from #60 since that gives an exception.

### Steps to reproduce
Pseudo code for what produces the log:

```python
options = {... Credentials(user, password) ...}
for product in list_of_many_products:
    download_feature(product, options)
```

### Expected behavior
Some kind of recovery, or returning a failure, or throwing an exception.

### Environment
 - OS: Ubuntu 22.04
 - Package version: 0.2.3, but also CDSETool main branch with #83 on top.
 - Python version: 3.10

pjonsson:
This happened again last night using main branch just before the proxy PR.
```
time=2024-03-07T22:27:51 level=INFO msg="Download started"
time=2024-03-07T22:27:52 level=WARNING msg="Status code 401, retrying.." file=download.py:104
time=2024-03-07T22:29:07 level=WARNING msg="Status code 401, retrying.." file=download.py:104
time=2024-03-07T22:30:08 level=WARNING msg="Status code 401, retrying.." file=download.py:104
```
@pedep a hanging job is a bit of an issue since that doesn't self-heal, any ideas?

pjonsson:
I thought the monitor had something to do with things, but they always seem to be cleaned up properly for me.

pedep:
I think i have an idea as to why the session will not refresh despite retrying for multiple seconds

```python
# [..snip..]
    with _get_monitor(options).status() as status:
        status.set_filename(filename)

        session = _get_credentials(options).get_session() # a session is acquired, which may be seconds from expiring
# [..snip..]
        response = _retry_backoff(url, session, options)
# [..snip..]

[...]

Issue #59, opened 2023-12-18
Return error indication from download_feature 


pjonsson:
## Summary

It would be nice to be able to distinguish success from failure when calling `download_feature`.

### Basic example

```python
if download_feature(..) is None:
   print("Failed to download")
```

### Motivation

The expected outcome is that success can be determined by the caller without performing file system operations and without digging around in data structures internal of CDSETool.

pedep:
Great suggestion :+1:

We will get around to this soon, unless someone wants to submit a PR? :)

PR #100, opened 2024-03-12
download: limit the number of retries

pjonsson:
The token will eventually expire, so
stop trying after 5 attempts failed.
This will result in a failure downloading
instead of a hanging job, so the root
cause remains, but the impact is reduced.

Refs #84

### DIFF
--- a/src/cdsetool/download.py
+++ b/src/cdsetool/download.py
@@ -39,26 +39,35 @@ def download_feature(feature, path, options=None):
         return filename
 
     with _get_monitor(options).status() as status:
+        (fd, tmp) = tempfile.mkstemp()  # pylint: disable=invalid-name
         status.set_filename(filename)
+        attempts = 0
+        while attempts < 5:
+            # Always get a new session, credentials might have expired.
+            session = _set_proxy(options, _get_credentials(options).get_session())
+            url = _follow_redirect(url, session)
+            with session.get(url, stream=True) as response:
+                if response.status_code != 200:
+                    log.warning(f"Status code {response.status_code}, retrying..")
+                    attempts += 1
+                    time.sleep(60 * (1 + (random.random() / 4)))
+                    continue
 
-        session = _get_credentials(options).get_session()
-        session = _set_proxy(options, session)  # here set the proxy
-        url = _follow_redirect(url, session)
-        response = _retry_backoff(url, session, options)
+                content_length = int(response.headers["Content-Length"])
 
-        content_length = int(response.headers["Content-Length"])
+                status.set_filesize(content_length)
 
-        status.set_filesize(content_length)
+                with open(fd, "wb") as file:
+                    for chunk in response.iter_content(chunk_size=1024 * 1024 * 5):
+                        file.write(chunk)
+                        status.add_progress(len(chunk))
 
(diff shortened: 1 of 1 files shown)


pjonsson:
Beyond limiting the number of attempts to 5, this PR preserves the existing logic and just moves code around which is a deliberate choice to lower the risk of introducing bugs. The choice of 5 attempts is arbitrary, it's much less than an infinite number of attempts, but more than immediately giving up.

I don't think the current values correspond to exactly what I will need in a month or two, but if anyone needs to fine-tune these values that can easily be done in a followup PR. As an added bonus, a PR that is just about fine-tuning values will be easy to review.

pedep:
Code looks fine as-is, but i will have to test it before merging.
The retry you refactored is not only used to catch 401 errors, but also the 429 errors returned by the error-prone rate-limiter used by copernicus.

Since its now possible for the retry request loop to give up, we should implement a way for the user to collect the errored. This could be a "resume file", or simply returning the ids of failed objects instead of `None`

pedep:
Could you deliberate on what you mean by this?
> moves code around which is a deliberate choice to lower the risk of introducing bugs.

pjonsson:
> The retry you refactored is not only used to catch 401 errors, but also the 429 errors returned by the error-prone rate-limiter used by copernicus.

Can you rephrase what you wrote? The expired key is never renewed in the current code, so after getting a 401, the download is stuck in that loop forever?

> Could you deliberate on what you mean by this?
>
> >  moves code around which is a deliberate choice to lower the risk of introducing bugs.

I resisted the temptation to try to make everything perfect, this PR is just a transplant of what is currently in the code, besides limiting the number of attempts.

pjonsson:
I'm not using the cli, but I've used this code for a couple of days and it does back out of the infinite loop when getting a 401 response.

I have never gotten a 429 that I'm aware of, but #99 should improve the situation with that since the Retry class respects Retry-After with its default parameters to the constructor. (I was not aware there was a rate limiter, so if the rate limiter is completely broken #99 will obviously not help.)

pedep:
> The retry you refactored is not only used to catch 401 errors, but also the 429 errors returned by the error-prone rate-limiter used by copernicus.

The download endpoint may return 429 errors in some edge cases. You can reproduce this by force restarting a multi-threaded download a few times in the span of one minute. The rate limiter will still have memory of your previous connections, and return 429 to your current connections. If you increase the concurrency above your account limit (regular accounts have a limit of 4) you can force the issue.

> I resisted the temptation to try to make everything perfect, this PR is just a transplant of what is currently in the code, besides limiting the number of attempts.

Thats fair.
I can see you also changed enough to fix the issue mentioned here https://github.com/SDFIdk/CDSETool/issues/84#issuecomment-1992083460 :+1:

pedep (approved):
I will refrain from requesting the change to the return value, as i can see we also return `None` in a few other places.

pjonsson:
> I will refrain from requesting the change to the return value, as i can see we also return `None` in a few other places.

Thanks! I'll mention that returning an ID when the download fails will unfix #59, but 
I'm guessing `download_features` has enough information to correlate what was passed in to `download_feature` and what was returned.

PR #101, opened 2024-03-12
implement download result dataclasses

pedep:
Allow users to gleam a bit more about the download result via result classes

breaking change in relation to #59


pjonsson:
What is the intended programming interface, code performs `isinstance` on the return value from `download_feature` to determine success/failure?

src/cdsetool/download.py
```diff
@@ -10,33 +10,76 @@
...
+
+
+@dataclass
+class DownloadFailure(DownloadResult):
+    """
+    Result class for failed downloads
+    """
+
+    def __init__(self, feature, message):
+        self.feature = feature
```
pjonsson, line 50: Can `feature` be in the base class? And the class only uses the id part of the feature, so perhaps id should be stored directly?

src/cdsetool/download.py
```diff
@@ -10,33 +10,76 @@
...
+        log.debug(f"Bad title ('{title}')")
+        return DownloadFailure(feature, "Feature has no title")
 
     filename = title.replace(".SAFE", ".zip")
     result_path = os.path.join(path, filename)
 
     if not options.get("overwrite_existing", False) and os.path.exists(result_path):
         log.debug(f"File {result_path} already exists, skipping..")
-        return filename
+        return DownloadSuccess(feature, result_path)
```
pjonsson, line 82: Returning `result_path` forces the caller to "subtract" the input argument `path` from the result to figure out the filename that was just downloaded, is that intended?

src/cdsetool/download.py
pedep: I agree we should store feature in the base class, but my intention was to allow the user full access to the feature object.

src/cdsetool/download.py
```diff
@@ -10,33 +10,76 @@
...
+
+
+@dataclass
+class DownloadSuccess(DownloadResult):
+    """
+    Result class for successful downloads
+    """
+
+    def __init__(self, feature, filename):
+        self.feature = feature
```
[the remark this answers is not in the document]
pedep: Very true. This is an oversight on my part.
I implemented these as regular classes, but swapped last moment to dataclasses.

pedep:
> What is the intended programming interface, code performs isinstance on the return value from download_feature to determine success/failure?

Yes, something along those lines. I am unsure if there is a more pythonic way of doing this, since `isinstance` is not nearly as clean as other languages allow for

eg. ruby switch/case has the `isinstance` feature built in
```ruby
res = download_feature(...)
case res
when DownloadSuccess
  puts("success")
when DownloadFailure
  puts("failure")
else
  puts("unknown")
end

File: src/cdsetool/download.py (hunk 2 above)
Comment (pedep) [reply src/cdsetool/download.py:79]: While it was intentional, i see that users who relied on the previous behaviour will have a difficult time with this solution.
I will remediate this

**Comment by pjonsson** [comment]
> > What is the intended programming interface, code performs isinstance on the return value from download_feature to determine success/failure?
> 
> Yes, something along those lines. I am unsure if there is a more pythonic way of doing this, since `isinstance` is not nearly as clean as other languages allow for

As a general observation, it feels like this PR is trying to solve a UX issue through means that you state are clumsy to interface with in Python. The caller of `download_feature' uses code that roughly looks like this:
```python
filename = download_feature(feature)
if filename is None
   print(feature.get("id")) # Or do anything else that uses feature.
   return False
```
so I have a hard time to see why the feature needs to be returned on failure. But if that is a requirement, a return type of a tuple of a boolean and a string (the pair(Success, Filename), or the pair(Fail, FeatureID)) would be easier to work with on the caller side.

I haven't used `download_features`, but I assume the returned list is unordered, so `result[1]` does not necessarily correspond to `features[1]`, and then `download_features` needs to track the information about which feature index the first result corresponds to.

src/cdsetool/download.py
pedep: I think some users might enjoy also having the absolute path to the download directory as a separate attribute in the `DownloadSuccess` struct. What are your thoughts on this?

pedep:
> I haven't used download_features, but I assume the returned list is unordered, so result[1] does not necessarily correspond to features[1], and then download_features needs to track the information about which feature index the first result corresponds to.

This is exactly the use-case for this feature. The majority of users i have interacted with strictly use `download_features` due to its multi-threading.

> a return type of a tuple of a boolean and a string (the pair(Success, Filename), or the pair(Fail, FeatureID)) would be easier to work with on the caller side.

While the `status, payload` tuple would be easier to work with, it has the downside of being less flexible when adding and or modifying in the future.
I could see a pattern like `(status, DownloadResult)` working out well, to circumvent the issue with `isinstance`. It allows for extensibility in the future, and ease of use due to `status` being a simple bool

pjonsson:
> Alternatively, the subclasses can be removed, and the status be expressed as a boolean on the `DownloadResult` class.

This is Python 3.10, but I think the single class could be easy to work with:
```python
@dataclass
class DownloadResult:
   feature: dict[str, Any]
   filename: str | None
   message: str | None
```
and on the calling side:
```python
rv = download_feature(feature)
if rv.filename is None:
   print(feature.get("id"))
   return False
```

Since we're on the subject of concurrency and you mentioned a faulty rate-limiter, why is the number of worker threads `1.5 * concurrency` instead of just `concurrency`? And couldn't the entire pool-implementation be simplified by submitting all futures at once and using `concurrent.futures.as_completed()` instead of the watermark+wait+list append?

pedep:
> Since we're on the subject of concurrency and you mentioned a faulty rate-limiter, why is the number of worker threads 1.5 * concurrency instead of just concurrency? And couldn't the entire pool-implementation be simplified by submitting all futures at once and using concurrent.futures.as_completed() instead of the watermark+wait+list append?

> why is the number of worker threads 1.5 * concurrency instead of just concurrency

https://github.com/SDFIdk/CDSETool/blob/815740692299805d405aa78191d15d7dd2404822/src/cdsetool/_processing.py#L17
Low water mark is when new items are taken from the iterable, not the concurrency size. This could in principle be an arbitrary number >= concurrency.
I chose a bit of overhead because threads might finish working before more work is added to the queue. This is based on the assumption that worker threads may consume from the executor pool when the main thread is busy.

> And couldn't the entire pool-implementation be simplified by submitting all futures at once

This is not an option currently, since it would remove support for lazy generators such as `FeatureQuery`. Due to Copernicus using limit/offset based pagination, the request times can get quite high when running some queries. I have clocked queries of over 60 seconds when testing.
https://github.com/SDFIdk/CDSETool/blob/815740692299805d405aa78191d15d7dd2404822/src/cdsetool/query.py#L34-L72

pedep:
> Having these comments as code comments in _processing.py would be awesome, most of what you wrote is impossible to guess from reading the code.

You are right, this should be explained. I will use this PR as an opportunity to add comments too.

>I'm also curious about this part which submits one (?) job to the queue for each round in the loop:
```python3
            if len(futures) < low_water_mark:
                submit_item()
```
>Is there a guarantee that len(done) < workers, or will a download of 20 products start with 6 submitted futures, and then gradually scale down to a single future submitted?

The blocking `wait` call resumes after just 1 future has resolved. This essentially clamps `assert len(done) == 1`.

https://github.com/python/cpython/blob/415cd06d724762f23b42f1ab36867b8114714684/Lib/concurrent/futures/_base.py#L96-L109

This should guarantee the "one out, one in" pattern.

If we decide not to trust the concurrent system, a very quick safety could be changing

            if len(futures) < low_water_mark:
                submit_item()
to

            while len(futures) < low_water_mark:
                submit_item()

pjonsson:
> > Is there a guarantee that len(done) < workers, or will a download of 20 products start with 6 submitted futures, and then gradually scale down to a single future submitted?
> 
> The blocking `wait` call resumes after just 1 future has resolved. This essentially clamps `assert len(done) == 1`.
> 
> https://github.com/python/cpython/blob/415cd06d724762f23b42f1ab36867b8114714684/Lib/concurrent/futures/_base.py#L96-L109
> 
> This should guarantee the "one out, one in" pattern.

I was looking at the documentation (https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.wait) which says:

> The first set, named done, contains the futures that completed (finished or cancelled futures) before the wait completed. The second set, named not_done, contains the futures that did not complete (pending or running futures).

so I don't see anything about the interface guaranteeing there is a single future returned. If we're heading into the implementation, can't:

https://github.com/python/cpython/blob/415cd06d724762f23b42f1ab36867b8114714684/Lib/concurrent/futures/_base.py#L289-L294

be triggered and return multiple items in the `done` set?

pedep:
It seems are right. I did a quick test using this bit of code, and get quite inconsistent results.

```python3
from concurrent.futures import wait, FIRST_COMPLETED
from concurrent.futures import ThreadPoolExecutor
import time

def fn(n):
    time.sleep(n)
    print(n)

    return n

with ThreadPoolExecutor(max_workers=20) as executor:
    not_done = [
        executor.submit(fn, 1),
        executor.submit(fn, 1),
        executor.submit(fn, 1),
        executor.submit(fn, 1),
        executor.submit(fn, 2),
        executor.submit(fn, 2),
        executor.submit(fn, 2),
        executor.submit(fn, 2),
    ]

    print('waiting for any')

    while len(not_done) > 0:
        done, not_done = wait(not_done, return_when=FIRST_COMPLETED)
        not_done = list(not_done)
        print('done:', len(done), 'not_done:', len(not_done))
```

The order in which the futures complete, and whether or not 1 or more futures are included in the `done` set, is entirely dependent on the python runtime and os scheduler's whims.

At least as long as the futures complete at the exact same time.

If the futures complete just a single ms or less after each other, they are waited and returned one by one every time.

```python3
        executor.submit(fn, 1),
        executor.submit(fn, 1.001),
        executor.submit(fn, 1.002),
        executor.submit(fn, 1.003),
        executor.submit(fn, 2),
        executor.submit(fn, 2.001),
        executor.submit(fn, 2.002),
        executor.submit(fn, 2.003),
```
 
Not that this excuses it, but it might explain why we have not noticed any bad behaviour in this regard, since the likelyhood of more than one of the downloads completing within < 1ms of quite low.

I will implement the `while` loop fix as suggested above

src/cdsetool/download.py
```diff
@@ -10,33 +10,64 @@
...
+class DownloadResult:
+    """
+    Describes the result of a download operation, whether it was successful or not.
+    Contains relevant information about the feature and the download operation.
+    """
+
+    success: bool
+    feature: dict
+    filename: str
+    message: str
```
pjonsson, line 32: How about:
```python
feature: dict
filename: Optional[str]
message: Optional[str]

@staticmethod
def ok(feature, filename):
    return DownloadResult(feature, filename, None)

@staticmethod
def fail(feature, message):
    return DownloadResult(feature, None, message)
```
with an invariant that either filename or message is None? Object construction is then done with:
```python
return DownloadResult.ok(feature, filename)
```

I'm not sure about having just `dict`, aren't all keys strings so the type can be `dict[str, Any]`?

pedep: I like the static contructor methods, but are you sure about using `filename` and `message` as indicators for  success?
Having the extra bool field is much easier to understand, for basically no extra code.

pedep: [491ebec](https://github.com/SDFIdk/CDSETool/pull/101/commits/491ebecb23f33236af7e03e6b5445b2e4f7d2543)

pjonsson: I have only used mypy, but I assume all the type checkers will also require people to write code like this:
```python
rv = download_feature(feature, options)
if not rv.success or rv.filename is None:
   return None
unzip(rv.filename)
```
since the type checker can't figure out the invariant that `success = True <=> filename != None`.

I don't know Python, but is this the use-case for property, so:
```python
@property
def success(self):
     return filename is not None
```
should be used?

pedep: Assuming the python type checkers are as smart as the TypeScript ones, they should be able to figure out the invariants.
I will configure one or two and have a go at it before i move forward with this

pjonsson: I think the property solution would bring the best of both worlds, no derived data stored in the object but it's still possible to do the status check with `rv.success`.

(I've been thinking about suggesting type annotations because I think it would be a good idea, but .. too many other things that also needs to be fixed)

pjonsson:
@pedep seen this, but need to do other things right now. Will take a look at this either tomorrow or Wednesday.

pedep:
@pjonsson Please take all the time you need :+1:

src/cdsetool/_download_result.py
```diff
@@ -0,0 +1,39 @@
...
+    Contains relevant information about the feature and the download operation.
+    """
+
+    success: bool
+    feature: Dict[str, Any]
+    filename: Optional[str]
+    message: Optional[str]
+
+    @staticmethod
+    def ok(feature, filename):
```
pjonsson, line 22: The functions should have type signatures.

Issue #74, opened 2024-02-02
Use of a proxy in internet calls


LattaruoloAndrea:
## Summary
Is it possible to add in the options the proxies for the internet requests? It could be useful in some circumstances 

### Basic example
```Python
list(
    download_features(
        features,
        "path/to/output/folder/",
        {
            "concurrency": 4,
            "monitor": StatusMonitor(),
            "credentials": Credentials("username", "password"),
            "proxies": {"http":"http_proxy", "https": "https_proxy",...}
        },
    )
```
then on the session it could be done something like as mentioned in [stackoverflow](https://stackoverflow.com/questions/30837839/how-can-i-set-a-single-proxy-for-a-requests-session-object)
```Python
import requests
proxies = {'http': 'http://10.11.4.254:3128'} #this proxies we could take from the options
s = requests.session()
s.proxies.update(proxies)
s.get("http://www.example.com") 
```
### Contributions
I' ve been trying to create a new branch but I don't have the right permission to do so
![permissions](https://github.com/SDFIdk/CDSETool/assets/36859220/d40c050c-943b-40ad-88aa-103ebc020207)
. If possible I would like to contribute to the project. Thanks

pedep:
@LattaruoloAndrea This is a great suggestion. We will definitely put http/https proxies on the roadmap!

As for contributing, you should follow the steps laid out in the [contributing section of the readme](https://github.com/SDFIdk/CDSETool?tab=readme-ov-file#contributing) :nerd_face: 

For more in-depth guides, have a look at this documentation
- https://git-scm.com/book/en/v2/GitHub-Contributing-to-a-Project
- or https://docs.github.com/en/get-started/exploring-projects-on-github/contributing-to-a-project

If you have accidentally cloned the SDFIdk/CDSETool repository, and wish to change the upstream to your new fork, please have a look at this article
- https://admcpr.com/what-the-fork/

pedep:
closed by #102 and #92

PR #102, opened 2024-03-14
Move proxy to credentials

pjonsson:
Return a fully configured session
from Credentials.get_session(),
including the proxy settings.
This moves most of the proxy-related
things out of download into
Credentials, which is good
preparation for #93.

Refs #74


pedep (approved):
Looks good :+1:

PR #116, opened 2024-03-28
credentials: set precise expire times

pjonsson:
Use the exp/iat times in the token
to compute precise expire times.

### DIFF
--- a/src/cdsetool/credentials.py
+++ b/src/cdsetool/credentials.py
@@ -140,7 +140,6 @@ class Credentials:  # pylint: disable=too-few-public-methods disable=too-many-in
             ),
             proxies=self.__proxies,
         )
-        now = datetime.now()
         response = session.post(self.__token_endpoint, data=data, timeout=120)
 
         if response.status_code == 401:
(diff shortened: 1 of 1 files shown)


src/cdsetool/credentials.py
```diff
@@ -170,17 +168,22 @@ def __ensure_tokens(self):
...
                 raise TokenClientConnectionError from e
-            jwt.decode(
+            data = jwt.decode(
                 self.__access_token,
                 key=key.key,
                 algorithms=key._algorithms,  # pylint: disable=protected-access
                 options={"verify_aud": False},
             )
+            if refresh_expire_delta is not None:
+                self.__access_token_expires = datetime(*localtime(data["exp"])[:6])
```
pedep, line 183:
```suggestion
                self.__access_token_expires = datetime.fromtimestamp(data["exp"])
```

pedep:
```python3
ts = 1715611926
assert datetime(*localtime(ts)[:6]) == datetime.fromtimestamp(ts)

File: src/cdsetool/credentials.py [hunk 2]
```diff
@@ -170,17 +168,22 @@ def __ensure_tokens(self):
...
+                self.__refresh_token_expires = (
+                    datetime(*localtime(data["iat"])[:6]) + refresh_expire_delta
```
pedep, line 185:
```suggestion
                    datetime.fromtimestamp(data["iat"]) + refresh_expire_delta
```

pedep (changes requested):

src/cdsetool/credentials.py
pjonsson: Much better, thanks! Fixed!

src/cdsetool/credentials.py
pjonsson: Fixed.

pedep (approved):

PR #121, opened 2024-04-02
query: re-use session/retry logic from credentials

pjonsson:
The Copernicus web site sometimes responds 500 (along other error codes) to queries. Use the retry logic from credentials so queries are retried just like downloads.

This PR creates a merge conflict with #120 so one has to be merged before the other, but the changes are complementary.


pjonsson:
@pedep just rebased this. Have been running all the outstanding PRs for my downloads while you were gone and things have worked well locally.

pedep (approved):

PR #140, opened 2024-05-18
logger_test: allocate object before call

pjonsson:


tests/logger/logger_test.py
```diff
@@ -36,6 +36,6 @@ def test_noop_does_not_error():
             },
             "somewhere",
         )
-        NoopLogger.debug("NoopLogger did not raise an exception")
+        NoopLogger().debug("NoopLogger did not raise an exception")
```
pjonsson, line 39: Is the assumption that `test_noop_logger_is_default` has run before this test so the `MagicMock()` has done something?

I'm not sure if there are guarantees of test order in general with pytest, but even if there is, that kind of design breaks running just this test.

pedep: The test was not written with that intent, but a quick test seems to confirm your theory. As you say, this is not ideal..

I think this may be the easiest way to preserve the intended test functionality.
```patch
diff --git a/tests/logger/logger_test.py b/tests/logger/logger_test.py
index 4ad99bb..022ebe2 100644
--- a/tests/logger/logger_test.py
+++ b/tests/logger/logger_test.py
@@ -6,6 +6,7 @@ from cdsetool.download import download_feature


 def test_noop_logger_is_default():
+    src = NoopLogger.debug
     NoopLogger.debug = MagicMock()

     assert NoopLogger.debug.call_count == 0
@@ -23,6 +24,7 @@ def test_noop_logger_is_default():

     assert NoopLogger.debug.call_count == 1

+    NoopLogger.debug = src

 def test_noop_does_not_error():
     try:
```

We need to keep in mind that the patch will persist if the assertion fails. I don't think this will be a problem in practice, since a failed assertion will require intervention anyways.

pjonsson: I don't claim to understand Python classes, but here's something from the interpreter:
```python
>>> class G:
...   def test(self):
...     print("hello")
...
>>> G.test()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: G.test() missing 1 required positional argument: 'self'
>>> G.test(None)
hello
```
So as far as I can tell, the line this PR is changing passes a string as "self", and that apparently works at runtime, but it's a strange thing to do so pyright complaining about it seems reasonable.

Whether the mock should stick past a single test ends seems like an orthogonal question, I think the usual way to solve that kind of stuff with pytest is a fixture with a suitable scope (sesssion/file/test/etc).

pedep (approved):

PR #147, opened 2024-05-22
logger: align methods with Logger

pjonsson:
Use the method signatures of Logger
for the NoopLogger.


pjonsson:
There's no advanced logging going on so the mis-alignment probably doesn't matter, but I think this will improve the situation for a type checker.

pedep (approved):

PR #162, opened 2024-07-17
credentials: retry more to login

pjonsson:
I initially set the retry-value
to 2 to prevent spamming bad login
credentials at the Copernicus
site, but this value is too low
when the site responds with 503.

Bump the value to 15 so people
don't get an exception thrown
in their face because of intermittent
issues with the Copernicus site.


pjonsson:
@subnet001 trying to get rid of errors/warnings in my logs. I don't need a release with this fix, but could you please merge this to main so I don't have to install from a fork?

subnet001 (approved):

PR #164, opened 2024-07-17
download: close/flush before copy

pjonsson:
Make sure to close the file so all
buffers are flushed before copying
the file.

This fixes an issue where the last
kilobyte(s) are missing from the downloaded
file, resulting in a corrupt zip file.

Refs #67

### DIFF
--- a/src/cdsetool/download.py
+++ b/src/cdsetool/download.py
@@ -65,15 +65,17 @@ def download_feature(
                 log.warning("Token signature expired, retrying..")
                 continue
             url = _follow_redirect(url, session)
-            with session.get(url, stream=True) as response:
+            with session.get(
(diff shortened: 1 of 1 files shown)


pjonsson:
@subnet001 sorry for the noise today, but this fixes a corner case where I've been getting "randomly" broken zip files that was confusing to me. My frequency so far is 150/257343, but it's possible other users have been more unlucky.

I realize it's probably in the middle of vacation, but I think this fix warrants a new CDSETool release.

subnet001:
It's completely fine, if anything I'm sorry for not merging in your contributions fast enough. We really appreciate your effort!

subnet001 (approved):

Issue #150, opened 2024-05-27
Download on WINDOWS OS fails


LattaruoloAndrea:
## Describe the bug
Hello, I tried to run the simple code and test it on  two different windows machines and I had the same error. Basically there is a permission denied, once the script tries to read from the temporary directory to copy the  downloaded file. (I used the shell with administrative privileges)
```
Traceback (most recent call last):
  File "C:\Users\geouser\test_sentinel\test_sentinel_lib_no_proxy.py", line 24, in <module>
    list(
  File "C:\Users\geouser\test_sentinel\env\Lib\site-packages\cdsetool\download.py", line 114, in download_features
    yield from _concurrent_process(
  File "C:\Users\geouser\test_sentinel\env\Lib\site-packages\cdsetool\_processing.py", line 61, in _concurrent_process
    yield future.result()
          ^^^^^^^^^^^^^^^
  File "C:\ProgramData\miniconda3\Lib\concurrent\futures\_base.py", line 449, in result
    return self.__get_result()
           ^^^^^^^^^^^^^^^^^^^
  File "C:\ProgramData\miniconda3\Lib\concurrent\futures\_base.py", line 401, in __get_result
    raise self._exception
  File "C:\ProgramData\miniconda3\Lib\concurrent\futures\thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\geouser\test_sentinel\env\Lib\site-packages\cdsetool\download.py", line 112, in _download_feature
    return download_feature(feature, path, options)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\geouser\test_sentinel\env\Lib\site-packages\cdsetool\download.py", line 88, in download_feature
    shutil.copy(file.name, result_path)
  File "C:\ProgramData\miniconda3\Lib\shutil.py", line 419, in copy
    copyfile(src, dst, follow_symlinks=follow_symlinks)
  File "C:\ProgramData\miniconda3\Lib\shutil.py", line 256, in copyfile
    with open(src, 'rb') as fsrc:
         ^^^^^^^^^^^^^^^
PermissionError: [Errno 13] Permission denied: 'C:\\Users\\LOCAL_~1\\Temp\\tmp8zf1spgi'
``` 

### Steps to reproduce
My code is as follow:

```python
[...]

Issue #51, opened 2023-12-07
Create tempfiles on same filesystem as output directory.


pedep:
## Summary
Use tempfiles for downloading on the same filesystem as the target directory.

### Motivation
The current implementation uses `fd, tmp = tempfile.mkstemp()` to acquire a place to download to. This might lead to cross-filesystem
copying in cases where the system tempdir is not the same as the download directory.

By using in-directory tempfiles this might be avoided, though it will increase the risk of leaving garbage in the output directory when downloads are terminated abruptly.

Other approaches with more sophisticated approached to locating a proper tempdir might be worth investigating.

### Basic example

```diff
diff --git a/src/cdsetool/download.py b/src/cdsetool/download.py
index b8e355e..5f591e9 100644
--- a/src/cdsetool/download.py
+++ b/src/cdsetool/download.py
@@ -9,6 +9,7 @@ import random
 import tempfile
 import time
 import shutil
+import string
 from cdsetool._processing import _concurrent_process
 from cdsetool.credentials import Credentials
 from cdsetool.monitor import NoopMonitor
@@ -41,7 +42,7 @@ def download_feature(feature, path, options=None):
 
         status.set_filesize(content_length)
 
-        fd, tmp = tempfile.mkstemp()  # pylint: disable=invalid-name
+        tmp = temp_filename(path, filename)
         with open(tmp, "wb") as file:
             for chunk in response.iter_content(chunk_size=1024 * 1024 * 5):
                 if not chunk:
@@ -50,11 +51,13 @@ def download_feature(feature, path, options=None):
                 file.write(chunk)
                 status.add_progress(len(chunk))
 
-        os.close(fd)
         shutil.move(tmp, os.path.join(path, filename.replace(".SAFE", ".zip")))
 
     return feature.get("id")
 
+def temp_filename(path, filename):
+    suffix = "".join(random.choices(string.ascii_lowercase + string.digits, k=8))
+    return os.path.join(path, filename + "." + suffix)
 
 def download_features(features, path, options=None):
     """
```

pjonsson:
From a security standpoint, `temp_filename` is similar to `tmpnam(3)`, and my man page starts the description with:
```
DESCRIPTION
       Note:  avoid  using  these  functions; use mkstemp(3) or tmpfile(3) in‐
       stead.

```

Can't this be implemented by setting the `dir` parameter to `mkstemp`/`mkdtemp` to something suitable?

[...]

PR #168, opened 2024-07-23
Add temp directory (discussion windows  bug fixes)

LattaruoloAndrea:
Hello I have been working to check what was not working on windows, with the changes made on download.py now it should work. I also added a parameter to set the temp directory as well, if it's not needed we can remove it #150  should be fixed with this.


src/cdsetool/download.py
```diff
@@ -38,6 +38,7 @@ def download_feature(
     log = _get_logger(options)
     url = _get_feature_url(feature)
     title = feature.get("properties").get("title")
+    temp_dir = _get_temp_dir(options)
```
pjonsson, line 41: I think `temp_dir = options.get("tmpdir") or None` will give the result you want, and that should also give the nice behavior when the user specifies `"tmpdir": ""` in options.

pjonsson:
This looks like it implements my suggestion in #51 to be able to specify where to put the temporary directory (nice!), accidentally reverts my empty log-message fix, and changes something with the logic of temp file closing.

I think being allowed to specify the temporary directory location isn't controversial, could you please put that in a separate PR against latest main?

I don't understand the logic changes for file closing, but #164 fixed a bug where files weren't closed properly. Can you check if the logic change from this PR is still required on top of latest main?

src/cdsetool/download.py
LattaruoloAndrea: I tried to use the same logic as for log and url that it was created a function specifically but inside the function there is the logic that you mentioned.

LattaruoloAndrea:
I will try to create a different PR for the temp file, with the current implementations I found that there are a few things that I don't like.
```python
                tmp_file = os.path.join(temp_dir, "download.zip")
                with open(tmp_file, "wb") as file:
                    # Server might not send all bytes specified by the
                    # Content-Length header before closing connection.
                    # Log as a warning and try again.
                    try:
                        for chunk in response.iter_content(chunk_size=1024 * 1024 * 5):
                            file.write(chunk)
                            status.add_progress(len(chunk))
                    except (
                        ChunkedEncodingError,
                        ConnectionResetError,
                        ProtocolError,
                    ) as e:
                        log.warning(e)
                        continue
                # Close file before copy so all buffers are flushed.
                shutil.copy(tmp_file, result_path)
                return filename
```
Is it better to change the  `tmp_file = os.path.join(temp_dir, "download.zip")` with `tmp_file = tempfile.NamedTemporaryFile(dir=temp_dir, prefix=filename.replace(".zip", "")).name` since your implementations looks like that all the temp file will be having the same nave "download.zip" in the temp dir which in concurrent threads I guess that's an error, while `tempfile.NamedTemporaryFile()` creates a random Id and we add a prefix with the name of the file

pjonsson:
> Is it better to change the `tmp_file = os.path.join(temp_dir, "download.zip")` with `tmp_file = tempfile.NamedTemporaryFile(dir=temp_dir, prefix=filename.replace(".zip", "")).name` since your implementations looks like that all the temp file will be having the same name "download.zip" in the temp dir which in concurrent threads I guess that's an error, while `tempfile.NamedTemporaryFile()` creates a random Id and we add a prefix with the name of the file.

You get a unique temporary directory from each call to `TemporaryDirectory()`, so both main branch and this PR should be thread-safe.

> And then in the current implementations there is no logic to delete the tmp_file after it was copyied if I am not wrong. I guess if you manage to work on these I can try again on windows and check if it works.

The temporary directory and everything inside it is removed when the program leaves the with-statement that has the 'with tempfile.TemporaryDirectory() as ..:`.

LattaruoloAndrea:
>You get a unique temporary directory from each call to `TemporaryDirectory()`, so both main branch and this PR should be thread-safe.

What if you specify the dir in TemporaryDirectory() will it creates different directories inside the dir specified?

> The temporary directory and everything inside it is removed when the program leaves the with-statement that has the `with tempfile.TemporaryDirectory() as ..:`.
Great! So I can test it on windows basically

pjonsson:
> > You get a unique temporary directory from each call to `TemporaryDirectory()`, so both main branch and this PR should be thread-safe.
> 
> What if you specify the dir in TemporaryDirectory() will it creates different directories inside the dir specified?

Yes!

pjonsson (approved):

LattaruoloAndrea:
I just added the prefix with filename inside the tempDir so it get's easier to track what it is being downloaded. I guess it is better to leave it in the dirTemp instead of the file name.

subnet001 (approved):

PR #178, opened 2024-09-12
download: reformat with latest black

pjonsson:


pjonsson:
@subnet001 Can you merge this with broken tests? I think this and then #179 on top of that should make main green again.

subnet001 (approved):

PR #179, opened 2024-09-12
Drop Python 3.8 support

pjonsson:
The CI tests fail for Python 3.8
because of some changes in Fiona
and/or Geopandas. The Python 3.8
security support ends 31 October 2024.

Rather than spending time on fixing
something that will be obsolete
in a little more than a month,
drop the support for Python 3.8 now,
so the CI isn't constantly red.


subnet001:
Seems like the linter is failing. Can you reformat before merging? Thanks!

subnet001 (approved):

PR #184, opened 2024-09-20
Append .zip to .SAFE instead of replacing it. Treat Sentinel 3 files accordingly

### LINKED ISSUE #180: Refactor .SAFE with .zip replacement
## Summary
In https://github.com/CDSETool/CDSETool/blob/main/src/cdsetool/download.py#L47
a replacement of .SAFE with .zip is performed. While this is true for e.g. Sentinel 2, it is not valid for Sentinel 3 (which "title" ends with SEN3).
Moreover, a simple refactor could be performed for not replacing .SAFE with .zip, but appending .zip in the end, in order to maintain uniformity with the Copernicus Browser download behavior, which maintains the .SAFE in the filename

### Motivation
Maintain uniformity along pipelines

fbalaban:
Append .zip to .SAFE (for Sentinel 1 and 2).
Include a simple if for Sentinel 3 (which has a suffix of SEN3) to append .zip there also.

Closes #180

### DIFF
--- a/src/cdsetool/download.py
+++ b/src/cdsetool/download.py
@@ -44,7 +44,9 @@ def download_feature(
         log.debug(f"Bad URL ('{url}') or title ('{title}')")
         return None
 
-    filename = title.replace(".SAFE", ".zip")
+    filename = title.replace(".SAFE", ".SAFE.zip")
+    if title.split(".")[-1] == ".SEN3":
+        filename = title.replace(".SEN3", ".SEN3.zip")
     result_path = os.path.join(path, filename)
 
     if not options.get("overwrite_existing", False) and os.path.exists(result_path):
--- a/src/cdsetool/download.py
+++ b/src/cdsetool/download.py
@@ -44,9 +44,7 @@ def download_feature(
         log.debug(f"Bad URL ('{url}') or title ('{title}')")
         return None
 
-    filename = title.replace(".SAFE", ".SAFE.zip")
-    if title.split(".")[-1] == ".SEN3":
-        filename = title.replace(".SEN3", ".SEN3.zip")
+    filename = title + ".zip"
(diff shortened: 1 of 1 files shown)


src/cdsetool/download.py
```diff
@@ -44,7 +44,9 @@ def download_feature(
         log.debug(f"Bad URL ('{url}') or title ('{title}')")
         return None
 
-    filename = title.replace(".SAFE", ".zip")
+    filename = title.replace(".SAFE", ".SAFE.zip")
```
pjonsson, line 47: Does it work if you make this line `filename += ".zip"` instead?

fbalaban: Sure thing. However, `filename` is a new variable, which is created through `title`. 
So it would have to be `filename = title + ".zip"`

I have made these changes so not to change the coding style. But it can be done as per your proposal, no problem.

pjonsson: Sorry, I was in the middle of something else when making my comment. You're right, `filename = title + ".zip"` is what I had in mind, and should have written.

The reason I'm suggesting to just append the `".zip"` instead of having an exhaustive list of collections is that they keep making more data available from Copernicus over time so the list needs to be kept up to date. I know they recently added Sentinel-3 OLCI products, and apparently they have been providing Landsat data for some time already.

fbalaban: Sure, no problem. If this the only comment, I will proceed with the change.
For sure it's much cleaner and lighter. Do you have in mind any cases which this append would not be sufficient?

pjonsson: Well, it's hardly any guarantee, but I can't think of any issues. We'll wait and see what the people who can merge things say.

subnet001 (approved):

PR #185, opened 2024-09-20
Use Ruff instead of black

pjonsson:
This replaces black with Ruff.
The coding style of Ruff is based
on black so there are no significant
differences as evidenced by the size
of the diff of this PR. The main
difference is that Ruff is significantly
faster.

Ruff also provides the functionality
of isort and Flake8 with various plugins,
but this PR just replaces the formatting
parts done by black, any further linting
can be done in a future PR.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -3,16 +3,13 @@ name: Lint
 on: [push, pull_request]
 
 jobs:
-  black:
+  ruff:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - uses: actions/setup-python@v5
+      - uses: chartboost/ruff-action@v1
         with:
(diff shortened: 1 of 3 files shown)


pjonsson:
I have used Ruff since v0.3.4 in late March, and the experience has been great, it's both reliable and quick. This PR only enables the formatting parts of Ruff, but the Ruff linter found real bugs in the code bases I worked on despite other linters already being used.

The people behind Ruff are also responsive to bug reports, I reported some race conditions with the general vagueness that you usually have for those types of problems ("CI ran this and then got these error messages, but it has only happened once in hundreds of runs") , and from that vague report they had made a PR that fixed it within 12 hours from my report.

subnet001 (approved):

subnet001 (approved):

PR #196, opened 2024-10-14
Enable Ruff "I" lint category

pjonsson:
This gives deterministic order
of imports, which increases the chance
that Github will be able to merge
independent pull requests without
manual intervention.


pjonsson:
This is lower priority than #198, so please merge that before this one.

subnet001 (approved):

PR #198, opened 2024-10-14
Improve behavior when running under Python -O

pjonsson:
Move some code around so the code will work better under Python -O (where asserts are removed) when bad parameters are passed.

This will conflict with #196, I prefer to have this merged first.


pjonsson:
@subnet001 I need to make more changes in this area, could you please take a look at this PR?

subnet001 (approved):

PR #219, opened 2024-12-05
Added checksum feature

LattaruoloAndrea:
Hello I added the checksum feature that can be disabled in the options as 
`{'checksum': False}`
inside the options of the downloads. Whenever a checksum fails it deletes the downloaded temp file and starts from the beginning for the number of iterations. If it fails all the times then it deletes the corrupted file.

### DIFF
--- a/src/cdsetool/download.py
+++ b/src/cdsetool/download.py
@@ -199,7 +199,7 @@ def checksum_for_file(
     return False
 
 
-def get_expected_checksum(algo, checksums_list):
+def get_expected_checksum(algo: str, checksums_list: list[dict]):
     """Return a Value checksum for an algo among a list of dict"""
     for checksum in checksums_list:
         algorithm = checksum.get("Algorithm", "")


src/cdsetool/download.py
```diff
@@ -97,6 +99,10 @@ def download_feature(
                         log.warning(e)
                         continue
                 # Close file before copy so all buffers are flushed.
+                if checksum:
+                    checksum_for_file(tmp_file, title, options.get("proxies", None))
+                    os.remove(tmp_file)
+                    continue  # download retry
```
pjonsson, line 105: I've had plenty of archives containing files with bad checksums from Copernicus, but it's always been the source archives that have been faulty, so downloading them again will just give the same error again.

src/cdsetool/download.py
```diff
@@ -158,3 +164,31 @@ def _get_credentials(options: Dict) -> Credentials:
...
+    for value in array_values:
+        if value["name"] != file_name:
+            continue
+        checksums: list = value["Checksum"]
+        for checksum in checksums:
+            if checksum.get("Algorithm", "") == "MD5":
+                retrieved_value = checksum.get("Value", "")
+                calculated_value = calculate_value_checksum(file_path, "MD5")
+                return retrieved_value == calculated_value
+    assert False, f"Failed {file_name} odata does not contain any available checksum"
```
pjonsson, line 187: There are some asserts in the code from before that are intended to work as a way to stop the program. The asserts don't work when running `python -O`, so we should not introduce new asserts that are essential for the program working.

I think you can just return False here to signal that the checksum check  failed.

src/cdsetool/download.py
```diff
@@ -158,3 +164,31 @@ def _get_credentials(options: Dict) -> Credentials:
...
+    array_values = odata_for_checksum.get("value", [])
+    if len(array_values) == 0:
+        # no data available for checksum
+        return False
+    for value in array_values:
+        if value["name"] != file_name:
+            continue
+        checksums: list = value["Checksum"]
+        for checksum in checksums:
+            if checksum.get("Algorithm", "") == "MD5":
```
pjonsson, line 183: Does something like the following work instead of the current body of the loop?
```python
            match checksum.get("Algorithm"):
                case "MD5":
                    calculated_sum = hashlib.md5(file_path.read()).hexdigest()
                case "SHA3-256":
                    calculated_sum = hashlib.sha3_256(file_path.read()).hexdigest()
                case algorithm:  # pragma: no cover
                    raise ValueError(f"Unexpected checksum algorithm '{algorithm}'")
            expected_sum = checksum.get("Value", "")
            if calculated_sum != expected_sum:
                log.warning(f"Expected checksum {expected_sum} but got {calculated_sum} for {file_path}")
               return False
````

File: src/cdsetool/download.py [hunk 4]
```diff
@@ -97,6 +99,10 @@ def download_feature(
                         log.warning(e)
                         continue
                 # Close file before copy so all buffers are flushed.
+                if checksum:
+                    checksum_for_file(tmp_file, title, options.get("proxies", None))
```
pjonsson, line 103: There is no check of the return value from this function, won't this code continue to retry to download until it runs of out retries?

src/cdsetool/query.py
```diff
@@ -324,3 +324,25 @@ def _get_describe_doc(
...
+    """Get odata for checksum given a product name"""
+    session = Credentials.make_session(None, False, Credentials.RETRIES, proxies)
+    attempts = 0
+    while attempts < 10:
+        attempts += 1
+        with session.get(
+            "https://catalogue.dataspace.copernicus.eu"
+            f"/odata/v1/Products?$filter=Name eq '{name}'"
+        ) as res:
+            if res.status_code >= 500:
```
pjonsson, line 339: They actually return 404 every now and then as well, and then a bit later the system is fine and the same request works. So I think the check needs to be != 200 here as well.

src/cdsetool/download.py
```diff
@@ -158,3 +164,31 @@ def _get_credentials(options: Dict) -> Credentials:
...
+
+
+def checksum_for_file(
+    file_path: str, file_name: str, proxy: Union[Dict[str, str], None] = None
+) -> bool:
+    """Evaluate the checksum of a file. If the file was downloaded correctly"""
+    odata_for_checksum = get_odata_by_name(file_name, proxy)
+    array_values = odata_for_checksum.get("value", [])
+    if len(array_values) == 0:
+        # no data available for checksum
```
pjonsson, line 176: Do you know if there are a lot of products missing checksums now so this return is something expected even when things "work"?

src/cdsetool/download.py
```diff
@@ -158,3 +164,31 @@ def _get_credentials(options: Dict) -> Credentials:
...
+        return False
+    for value in array_values:
```
pjonsson, line 178: The odata query is for a file name for the product we just have downloaded, so I would guess the response contains information for that product, but there's a loop here, are there other products in the response that we need to iterate over?

src/cdsetool/download.py
LattaruoloAndrea: From my side I don't know if Copernicus resolved all the problems regarding the checksums. I remember there where some issues on older products missing the checksum or having an incorrect one  but not sure if things are now resolved. I guess it's better to still check if that "value" field is present than not to check it.

src/cdsetool/download.py
LattaruoloAndrea: That's true I missed that part 
```python
if checksum:
    if not checksum_for_file(tmp_file, title, options.get("proxies", None)):
        os.remove(tmp_file)
        continue  # download retry
```
this should solve the problem

src/cdsetool/download.py
LattaruoloAndrea: Ok perfect

src/cdsetool/download.py
LattaruoloAndrea: Yeah I know that that is a big problem, from Copernicus not providing reliable metadata. I guess you can just bypass the checksum by giving it in the parameters for download to false  (`{'checksum': False}`). Maybe I can put the default checksum value to false and then if someone needs it can be change it to true to use it. Even though this logic has to be switched in the future  since in a working environment where the checksum provided are valid we should always check if the products where downloaded correctly. Let me know if you prefer the checksum disabled from the beginning or not

src/cdsetool/download.py
LattaruoloAndrea: Actually I discovered that the match\case statement is was introduced in Python3.10 so the tests on python 3.9 fails. So let me know if we use the if elif clause or mantain the match\case.

pjonsson: Oh, I have no idea what Python version that introduced something, sorry about that! Keep the if-then-else.

The code in your reply to my comment loops through the checksums until it finds an acceptable checksum. I don't trust the metadata, so I'm wondering if the logic should instead be to try checksums in order of algorithmic preference, so we define something like (pseudo-code):
```python
if not checksums:
  return True
for name, fun  in [("BLAKE", hashlib.blake), ("SHA3-256", hashlib.sha3_256), ("MD5", hashlib_md5)]:
  expected_sum = get_expected_checksum(algorithm, checksums)
  if expected_sum:
    with open(file_path, "rb") as file:
      calculated_sum = fun(file.read()).hexdigest()
      return calculated_sum == expected_sum
return False
```
so we get a failure if there is a checksum for our algorithm of choice that doesn't match the file we have downloaded, and we also get a failure if there are checksums for the archive but only with algorithms we don't support.

src/cdsetool/download.py
LattaruoloAndrea: @pedep what do you think? For me personally I would still like to have a checksum parameter which I can use to check if the download was correct. Maybe we can set the default for this checksum to false so the overall logic for the application is not changed.

pjonsson: I want the checksum too, my question is whether re-downloading a product that fails the checksum validation will make any useful difference. Downloading the same broken product 10 times (and immediately discarding it because the checksum validation fails) will just consume bandwidth and eat into the Copernicus download quota for the user.

LattaruoloAndrea: I agree on what you are saying, but personally  I need to deal with this use case: I downloaded a product using two different machines on one machine I had the product downloaded without errors and on the other there where some data corrupted (I guess it is a problem with the internet connection that sometimes fails and it fetch data in a bad way  but this is just a wild guess). That's why the checksum would be useful. I agree that if the product that we are downloading is broken and/or the metadata are broken/missing  then it consumes bandwidth but in this case it's something that we cannot control. The only hope it would be that the metadata are correct and the products on Copernicus database are correct so that checksum would work accordingly.

src/cdsetool/download.py
LattaruoloAndrea: I implemented with the following

```python
        if not checksums:
            return True
        list_supported_checksum_algo = [("MD5", hashlib.md5)]
        # add new checksums algo here order list is relevant!
        for name, fun in list_supported_checksum_algo:
            expected_sum = get_expected_checksum(name, checksums)
            if expected_sum:
                with open(file_path, "rb") as file:
                    calculated_sum = fun(file.read()).hexdigest()
                    return calculated_sum == expected_sum
        return False
    log.warning(f"No valid checksum calculated for {file_name}")
    return False


def get_expected_checksum(algo, checksums_list):
    """Return a Value checksum for an algo among a list of dict"""
    for checksum in checksums_list:
        algorithm = checksum.get("Algorithm", "")
        if algorithm == algo:
            return checksum.get("Value", "")
    return ""
```

So now we have a list of supported algorithm, (I removed SHA-256 since I didn't see as the one used by Copernicus, maybe we could add BLAKE3) we loop through this list and we get the checksum starting from the first element of the list.
So now it loops just on the algorithm that we support and it gives a True result on the first correct checksum, if it doesn't find any it gives a False result.

src/cdsetool/download.py
pjonsson: Since transfers are done over TCP, I don't understand how corruption can happen in the "transfer" unless we have a bug in our code somewhere (which is not impossible). Was this something that happened recently? I fixed some product corruption issues this autumn.

LattaruoloAndrea: I noticed this at the start of December. For example what if the connection is not stable and sometimes the Pc disconnect from the internet. That's the case I was working with

src/cdsetool/download.py
pjonsson: I have (external) code that goes through the manifest in the product and validates the files in the product against the checksums in the manifest. I think "algorithms" in the manifest used to be MD5, but they changed to SHA3-256 somewhere along the way. I wonder if the product checksums are calculated in a completely different codepath compared to what generates the manifest..

MD5 has had known weaknesses for a long time by now and it seems BLAKE3 isn't supported by hashlib even in the most recent Python. Does the metadata contain checksums for any other algorithm than MD5 that hashlib in Python 3.9 still supports (https://docs.python.org/3.9/library/hashlib.html#hash-algorithms)?

src/cdsetool/download.py
pjonsson: Just to be clear, I'm not at all doubting your observations, I just don't understand how they can happen.

The disconnecting scenario that you describe will mean the other end starts sending packets 1-100, your PC gets packets 1-20 and then disconnects for a bit before it comes back and gets packets 40-50. At that point TCP will detect that packets 21-39 did not arrive and resend them. It's been a while since I looked at the retransmission mechanism in TCP so I'm sure the details in that simplified explanation are wrong, but it conveys the big picture of how it works.

If you were running the latest CDSETool in December you definitely had all the corruption fixes. Are both your PCs running Windows?

src/cdsetool/download.py
pjonsson: Ok, let's go with MD5 for now then.

src/cdsetool/download.py
LattaruoloAndrea: Yes the test on the products where executed on both windows machine, while on Ubunto I tested and never had this kinda of a problem. I know that with TCP it should work but so far the test game me two different result and it was the only thing that i could think of. Since they where on two different networks but with the same code and using the same list of pruducts. It's what I posted on the issue [128](https://github.com/CDSETool/CDSETool/issues/218), which led to the checksum implementation

src/cdsetool/download.py
subnet001: I'm ready to resolve this conversation this but I'm a bit out of my depth here, so I'll defer to @pedep

pjonsson: Here's the weakness I was talking about: https://www.kb.cert.org/vuls/id/836068/

If someone targets a user and manages to forge the data they download, there is a possibility the MD5 checksum will not catch it. But compared to not checking the checksum at all as we do right now, this PR is still a massive improvement.

src/cdsetool/download.py
```diff
@@ -158,3 +166,43 @@ def _get_credentials(options: Dict) -> Credentials:
...
+            if expected_sum:
+                with open(file_path, "rb") as file:
+                    calculated_sum = fun(file.read()).hexdigest()
+                    return calculated_sum == expected_sum
+        return False
+    log.warning(f"No valid checksum calculated for {file_name}")
+    return False
+
+
+def get_expected_checksum(algo, checksums_list):
```
pjonsson, line 202: Please add a return type for the function and the types for the parameters here.

LattaruoloAndrea: I should have fixed the missing params

src/cdsetool/download.py
massimozanetti: > I've had plenty of archives containing files with bad checksums from Copernicus, but it's always been the source archives that have been faulty, so downloading them again will just give the same error again.

Hi @pjonsson I'have been trying checksums for some products and I see them ok. Can you make some examples of bad checksums so I can try?

pjonsson: I haven't run this code, but for testing the external checksum code I mentioned I changed the checksum comparison from `==` to `!=` to trigger a failure.

Be careful with doing that test on more than a single product though (since the check will repeatedly fail and all specified products will be downloaded and immediately discarded 10 times).

Issue #131, opened 2024-05-01
Download specific file through node filtering


tvoirand:
Hi,

Thanks a lot for your work on this project!

## Summary

Sentinelsat allowed to download a specific file within a product bundle using a "node filtering" functionality. I think it would be great if CDSETool also provided this. 

I had a very brief look at the CDSE documentation, it seems the node functionality is available through their [OData API](https://documentation.dataspace.copernicus.eu/APIs/OData.html#listing-product-nodes), and not through the [OpenSearch API]() that CDSETool is using, so it might not be that straightforward to add. Anyway I'd be curious to know your opinion!

### Basic example

The functionality I'm referring to in sentinelsat is described [here](https://sentinelsat.readthedocs.io/en/stable/api_overview.html#downloading-parts-of-products).
 
### Motivation

It would reduce the download size for users interested in a single or a few files within the product.

tvoirand:
Hello,

For information, I have implemented this functionality and I am using it within a personal project.

If maintainers of this repo think this has any interest, I would be happy to open a PR, and adapt the code to generic use cases if necessary.

Cheers!

fbalaban:
Hi @tvoirand, this is indeed interesting and we also want to include a feature of tile list filtering.
Could you please create a fork from the latest version of CDSETool, showing the functionality you have implemented? Then, I would be happy to contribute and extend your implementation, so to propose to the maintainers to merge it.
Thanks!

tvoirand:
Hi @fbalaban,

Thanks for your interest! The solution I am using so far is available [here](https://github.com/tvoirand/CDSETool/blob/nodefilter/src/cdsetool/nodefilter.py). It can be used as follows:
```python
from cdsetool.query import query_features
from cdsetool.nodefilter import download_node

features = list(
    query_features(
        "Sentinel2",
        {
            "startDate": "2020-01-01",
            "completionDate": "2020-01-04",
            "processingLevel": "S2MSI2A",
            "box": ["4", "51", "4.5", "52"],
        },
    )
)

for feature in features:
    download_node(
        feature.get("id"),
        path="XXX",
        username="XXX",
        password="XXX",
        nodefilter_pattern="*TCI_60m.jp2",
    )
```

This solution consists in walking the product's nodes tree, listing every file matching the filter pattern, and downloading the listed files. This is sufficient for my use case, but to make it a proper feature of CDSETool I would see a couple of minimal improvements:
[...]

PR #228, opened 2025-01-05
Enable filtering and selective file downloads via OData API

tvoirand:
### Summary

Hi,

As first mentioned in #131, I am proposing to add a functionality for filtering and downloading specific files within product bundles.

Filtering is based on Unix filename pattern matching, and downloading relies on OData API product nodes.

### Description of the proposed changes

- **`README.md`:** Add examples for downloading specific files within products bundles
- **`cli.py`:** Add `filter_pattern` option to the `cdsetool download` command
- **`download.py`:**
  - Add `filter_files` function to parse product metadata and filter files based on a given pattern
  - Add `_href_to_url` function to convert a file's `href` from the metadata into an OData download URL
  - Add `download_nodes` function to download specific files within a product bundle based on a given filename pattern string
  - Add `download_file` function to handle downloading a single file within a product bundle
  - Modify `download_feature` to call the new `download_nodes` function when a `filter_pattern` option is specified
- **`tests/download/`:** Add tests based on mock metadata files to validate the new functionality

### Other details

This is a first proposition, and if it raises interest, I guess it will need some improvements before merging. I was already thinking of the following points:
- Raise exceptions when download fails, instead of returning None
  - Makes it cleat that an error has occurred and that it needs to be handled by the caller
  - Reduces consistency if rest of the codebase relies on returning and handling None
- Zip the result
  - Improves consistency with download_feature
- Use checksum to check downloaded files integrity (but I see that's already mentioned in a dedicated PR)
- For now only Sentinel-1, 2, and 3 products are supported, this could be completed with other products provided by CDSE's OData API (for ex. Landsat-5, 7, and 8)
- I have a couple of pylint errors, that I think will prevent the CI pipeline from passing, but I wanted to wait for some feedback to fix this (notably a "too many arguments" error in `cli.py`, but maybe adding a dedicated command is preferable?)


src/cdsetool/download.py
```diff
@@ -26,6 +29,175 @@
...
+                path.split("/")[1:]
+            )  # Remove product name prefix in S3 manifests
+            match = fnmatch.fnmatch(path.lower(), pattern)
+            if match and not exclude or exclude and not match:
+                paths.append(path)
+
+    return paths
+
+
+def download_file(url: str, path: str, options: Dict[str, Any]) -> Union[str, None]:
```
pjonsson, line 85: I don't think anyone is against being able to download single files with CDSETool, the problem was just that nobody had implemented the functionality.

Looking at the implementation, this and the next function looks like they are more or less copies of some other function. I think it's better that you put this functionality into the existing functions under some if-else statement so most of the code is shared and we don't get even more places to maintain where we have to play whack-a-mole when we figure out new ways that the remote servers misbehave.

I haven't looked at the details of the code since integrating it into the existing functions is bound to change a bunch of things, but from a quick scan I think the code itself looks good.

src/cdsetool/download.py
```diff
@@ -26,6 +29,175 @@
 from cdsetool.monitor import NoopMonitor, StatusMonitor
 from cdsetool.query import FeatureQuery
 
+MANIFEST_BASENAMES = {
+    "SENTINEL-1": "manifest.safe",
+    "SENTINEL-2": "manifest.safe",
+    "SENTINEL-3": "manifest.xml",
```
pjonsson, line 35: What happens when someone tries to use this functionality for something that isn't Sentinel-1/2/3?

src/cdsetool/download.py
```diff
@@ -26,6 +29,175 @@
...
+    set to True, only files that do not match pattern are returned.
+    """
+    paths = []
+    xmldoc = etree.parse(manifest_file)
+
+    if os.path.basename(manifest_file) == "manifest.safe":
+        data_obj_section_elem = xmldoc.find("dataObjectSection")
+        for elem in data_obj_section_elem.iterfind("dataObject"):
+            path = elem.find("byteStream/fileLocation").attrib["href"]
+            path = path[2:]  # Remove "./" prefix present in S2 and S3 manifests
```
pjonsson, line 65: Why does `./` need to be removed, isn't that just the current directory?

src/cdsetool/download.py
tvoirand: Thanks for your prompt feedback! I will review this to avoid code duplication.

src/cdsetool/download.py
tvoirand: I haven’t properly managed this case so far, I will have a look it. Raising an exception is probably a good solution.

By the way other product types can be supported, as long as their metadata can be parsed by `filter_files` to read specific files paths within product bundles.

src/cdsetool/download.py
```diff
@@ -26,6 +29,175 @@
...
+            match = fnmatch.fnmatch(path.lower(), pattern)
```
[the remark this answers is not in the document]
tvoirand: Thanks for the catch!

src/cdsetool/download.py
```diff
@@ -26,6 +29,175 @@
...
+            if match and not exclude or exclude and not match:
```
pjonsson, line 67: This looks like xor, `if match ^ exclude:`.

The `match = fnmatch..` and if-statement is identical between the two blocks, so share it:
```python
            path = "/".join(...)
    else:
        return paths
    match = fnmatch....
```

tests/download/test_download.py
```diff
@@ -0,0 +1,213 @@
+"""Tests for CDSETool's download module."""
+
+import os
+import tempfile
+from unittest import mock
```
pjonsson, line 5: The credentials tests mock things with a different mechanism, see if it's possible to use that. You might be able to import those mocks directly.

tests/download/test_download.py
```diff
@@ -0,0 +1,213 @@
...
+    href = "path/to/resource.xml"
+    expected_url = (
+        "http://example.com/odata/v1/Products(a6215824-704b-46d7-a2ec-efea4e468668)/"
+        "Nodes(S2B_MSIL1C_20241209T162609_N0511_R040_T17UPV_20241209T195414.SAFE)/"
+        "Nodes(path)/Nodes(to)/Nodes(resource.xml)/$value"
+    )
+    assert _href_to_url(odata_url, product_id, product_name, href) == expected_url
+
+
+def test_filter_files_s1():
```
pjonsson, line 23: This and the next 3 tests look nearly identical, make it a single test and parameterize the manifest name, the filtered name, and the expected result. There's a quite similar example for `eval` here: https://docs.pytest.org/en/7.1.x/how-to/parametrize.html#parametrize-basics

tests/download/test_download.py
```diff
@@ -0,0 +1,213 @@
...
+    manifest_file_path = "tests/download/mock/sentinel_1/manifest.safe"
+    filtered_files = filter_files(manifest_file_path, "*/calibration-*.xml")
+    filtered_files.sort()
```
pjonsson, line 26: Make the expected output match the order things come out in instead of sorting (could mask bugs otherwise).

tests/download/test_download.py
```diff
@@ -0,0 +1,213 @@
...
+    mocker.patch(
+        "cdsetool.download.filter_files",
+        return_value=["./GRANULE/file1.jp2", "./GRANULE/file2.jp2"],
+    )
+    mocker.patch(
+        "cdsetool.download.download_file",
+        mock_download_file,
+    )
+
+    with tempfile.TemporaryDirectory(prefix="test_download_nodes_success") as final_dir:
```
pjonsson, line 171: Use the `tmp_path` fixture (https://docs.pytest.org/en/stable/how-to/tmp_path.html#the-tmp-path-fixture) instead of `TemporaryDirectory` in the tests.

src/cdsetool/download.py
```diff
@@ -26,6 +29,175 @@
...
+                paths.append(path)
+
+    elif os.path.basename(manifest_file) == "manifest.xml":
+        namespaces = {"ns": "http://www.eumetsat.int/sip"}
+        data_section_elem = xmldoc.find("ns:dataSection", namespaces)
+        for elem in data_section_elem.iterfind("ns:dataObject", namespaces):
+            path = elem.find("ns:path", namespaces).text
+            path = "/".join(
```
pjonsson, line 75: I'm not sure what exactly this does, but it looks like something that might already be present in pathlib (maybe https://docs.python.org/3.9/library/pathlib.html#pathlib.PurePath.joinpath, or some other operation).

src/cdsetool/download.py
pjonsson: I'm not sure how stable the manifest format is in practice, but even if it is completely fixed, I think it might make sense to use pathlib here instead of blindly cutting characters:
```python
>>> f = pathlib.Path("./test/directory/file.txt")
>>> f
PosixPath('test/directory/file.txt')
>>> str(f)
'test/directory/file.txt'
```
And it looks like the two callers you mentioned wouldn't mind getting a `Path` instead of a string.

pjonsson:
>     * I have a couple of pylint errors, that I think will prevent the CI pipeline from passing, but I wanted to wait for some feedback to fix this (notably a "too many arguments" error in `cli.py`, but maybe adding a dedicated command is preferable?)

The "code smell" warnings aren't always applicable, so prioritize writing good code that gives a good user experience. If that means PyLint complains, we can always change the limits or suppress the warning for that line/function.

src/cdsetool/download.py
subnet001: I agree with @pjonsson, i think its better to implement it into existing functions. I also think the functionality will be a very positive addition to CDSETool. 
Thank you for your contribution!

tests/download/test_download.py
tvoirand: Do you mean to compare both lists without sorting? Isn’t there some randomness in the order things come out, which is not necessarily related to bugs and which could lead the test to fail? If the list is sorted within filter_files like below, do you still see a risk of masking bugs?
```python
def filter_files(...):
    ...
    return sorted(paths)
```

src/cdsetool/download.py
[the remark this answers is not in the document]
pjonsson: One of the tests use `Oa02_reflectance` as pattern, and that looks like a band name to me. GDAL doesn't accept "kd490_m07" as name for the KD490_M07 band, so people working with this data must already be used to case-sensitivity. My guess is that long term, our users will benefit from the alignment with the GDAL behavior.

(I do have some first hand experience of being painted into a corner because someone previously ignored casing of band names, so I would like to do what I can to spare others from making the same mistake.)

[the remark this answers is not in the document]
tvoirand: Sure, this can work too with case-sensitivity, and thus removing both `lower()`.

I initally included this `lower()` because it was present in sentinelsat’s code that inspired this PR. But surely we don't need to reproduce exactly what sentinelsat was doing, and your argument makes sense to me.

Thanks for this advice, and all your other feedback. I am going to push my update once I go through all the points you mentioned (which might take some more time!).

src/cdsetool/download.py
tvoirand: Thanks, I can see how pathlib is useful in this particular case. But I haven't seen it used elsewhere in CDSETool, which is why I avoided it so far. For consistency, I think it would make sense to replace every use of `os.path` with pathlib, but maybe in a separate PR, if you agree!

src/cdsetool/download.py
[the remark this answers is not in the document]
pjonsson: I appreciate your contribution and there is absolutely no rush, I (and everybody else) understand that contributors usually have a day job that isn't about maintaining/extending the particular project that gets a pull request.

src/cdsetool/download.py
pjonsson: In this spot, I think using pathlib makes sense, and using pathlib doesn't add a dependency.

I haven't looked, but I'm not sure about replacing all other uses of `os.path` with pathlib since I think that also changes the type from `str` to `Path`. It's possible that some code should have used `Path` all along instead of `str`, but sometimes it just makes all callers add `Path(thing)` in the function argument and then the first thing inside the function is `str(thing)` to get the argument back into a string.

tests/download/test_download.py
tvoirand: Done with 14c3c5f.

tests/download/test_download.py
tvoirand: Understood, I updated the test to match the order things come out instead of sorting with 3324d74.

tests/download/test_download.py
tvoirand: Done with 773239f.

src/cdsetool/download.py
tvoirand: I added a try/except block to handle the case where the collection type is not supported with 6271303.

src/cdsetool/download.py
tvoirand: I have centralized the code with a41d92e, in a nutshell:
- I have removed the `download_nodes` function, and integrated its logic within `download_feature` under an if-statement
- I have kept `download_file` and updated `download_feature` to call it

Another thing to note:
- I have changed the temporary download file naming from `"download.zip"` to `f"{title}.zip"` to enable logging the title in case of failure in `download_file`


(And thank you subnet001 for the positive feedback!)

src/cdsetool/download.py
tvoirand: > that also changes the type from `str` to `Path`

That's also why I am not fully comfortable with having `filter_files` returning `Paths`, as it would result in using both `str` and `Path` in different parts of `download_feature`. 

Another alternative for `filter_files` would be to use pathlib locally and convert to `str` before returning. With the recent change to centralize the `fnmatch` call, this approach comes quite naturally. I am proposing this with commit dbe9361 if you want to have a look.

src/cdsetool/download.py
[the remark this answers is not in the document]
tvoirand: Done with 46109d3.

tests/download/test_download.py
tvoirand: I have just made an attempt to address your last pending comment regarding the mock mechanism used in credentials test. With a067cae and f5f84b9, this mock mechanism is centralized and used in both the credentials and download tests.

I hope the changes I made are clear and make sense. Please let me know if you’d like any clarifications or see other changes to make.

As a side note, I am considering adding a last comment describing the changes proposed in this PR. This can be useful for documentation, as the description in my initial comment when submitting the PR is not fully valid anymore. But I’ll wait until/if the PR is approved before doing so, in case things change again.

src/cdsetool/download.py
tvoirand: This is linked with the other discussion above, also on the topic of using paths in filter_files. I attempted to address this point with https://github.com/CDSETool/CDSETool/commit/dbe93616bae8deb331825b40a735fe861a4a15f2.

tvoirand:
> > ```
> > * I have a couple of pylint errors, that I think will prevent the CI pipeline from passing, but I wanted to wait for some feedback to fix this (notably a "too many arguments" error in `cli.py`, but maybe adding a dedicated command is preferable?)
> > ```
> 
> The "code smell" warnings aren't always applicable, so prioritize writing good code that gives a good user experience. If that means PyLint complains, we can always change the limits or suppress the warning for that line/function.

In this PR, the lint workflow raised 4 warnings. I am proposing to ignore them with 3f32da8 and e95321c, and I describe the reasons below. But this might be a little bold, so I would be happy to know your opinion.
- `download.py`
    - Pyright `reportOptionalMemberAccess` and `reportArgumentType` in `filter_files`:  These reports are caused by potential `None` values if something’s wrong with the manifest parsing. But fixing each Pyright report would involve adding many `if <variable> is not None:` checks, which would make the code more verbose, and exceptions in `filter_files` are handled by a try/except block in `download_feature`. Therefore, I propose to ignore these warnings.
    - Pylint `broad-exception-caught` in `download_feature`: If something’s wrong with the manifest parsing, an AttributeError will likely be raised (linked with the Pyright warning described above). However, other exception types could occur that I am not thinking of now, and they can be handled in the same way. Therefore, I propose to ignore this warning too.
    - Pylint `too-many-return-statements` in `download_feature`: These return statements come from handling errors by logging a message and returning None. Maybe they can be reduced by raising exceptions instead. But I propose to ignore this warning for now.
- `cli.py`
    - Pylint `too-many-arguments` and `too-many-positional-arguments` in `download`: This is caused by adding `filter_pattern` as a new argument, which surpasses the default limit (6/5). If this is still reasonable, we could ignore this warning or raise the limit.

src/cdsetool/download.py
```diff
@@ -1,16 +1,20 @@
...
+import fnmatch
 import os
 import random
 import shutil
 import tempfile
 import time
-from typing import Any, Dict, Generator, Union
+from pathlib import Path
+from typing import Any, Dict, Generator, List, Union
+from xml.etree import ElementTree as etree
```
pjonsson, line 17: The tutorial on python.org uses `from xml.etree import ElementTree as ET`, I think it makes sense to use the same name here.

src/cdsetool/download.py
```diff
@@ -26,31 +30,53 @@
...
+    elif os.path.basename(manifest_file) == "manifest.xml":
+        namespaces = {"ns": "http://www.eumetsat.int/sip"}
+        data_section_elem = xmldoc.find("ns:dataSection", namespaces)
+        for elem in data_section_elem.iterfind("ns:dataObject", namespaces):  # type: ignore
+            path = Path(elem.find("ns:path", namespaces).text)  # type: ignore
+            # Remove prefix present for some files in S3 manifests
+            path = Path(*path.parts[-1:])
+            paths.append(path)
+
+    paths = [str(path) for path in paths if fnmatch.fnmatch(path, pattern) ^ exclude]
```
pjonsson, line 67: Put the return on this line instead of redefining `paths`. The indentation level is 8 spaces on the line before so it's difficult to think this somehow belongs to the previous block, so I think the preceding blank line can be removed too.

src/cdsetool/download.py
```diff
@@ -96,13 +115,111 @@ def download_feature(
                     ) as e:
                         log.warning(e)
                         continue
-                # Close file before copy so all buffers are flushed.
-                shutil.copy(tmp_file, result_path)
-                return filename
+                return path
```
pjonsson, line 118: This "new" function now returns its input. Since callers know what they passed into the function, return a bool instead.

src/cdsetool/download.py
```diff
@@ -26,31 +30,53 @@
...
-        return None
+    if os.path.basename(manifest_file) == "manifest.safe":
+        data_obj_section_elem = xmldoc.find("dataObjectSection")
+        for elem in data_obj_section_elem.iterfind("dataObject"):  # type: ignore
+            path = Path(elem.find("byteStream/fileLocation").attrib["href"])  # type: ignore
+            paths.append(path)
 
-    filename = title + ".zip"
-    result_path = os.path.join(path, filename)
+    elif os.path.basename(manifest_file) == "manifest.xml":
```
pjonsson, line 58: I don't think there are any other places in the code that puts a blank line between the if and the else, so let's not start with that here.

src/cdsetool/download.py
```diff
@@ -26,31 +30,53 @@
...
-    temp_dir_usr = _get_temp_dir(options)
+    paths = []
+    xmldoc = etree.parse(manifest_file)
 
-    if not url or not title:
-        log.debug(f"Bad URL ('{url}') or title ('{title}')")
-        return None
+    if os.path.basename(manifest_file) == "manifest.safe":
+        data_obj_section_elem = xmldoc.find("dataObjectSection")
+        for elem in data_obj_section_elem.iterfind("dataObject"):  # type: ignore
```
pjonsson, line 54: Pyright elminates all type checking on the line with `type: ignore` (because the spec says so), use `pyright:ignore[ThingToIgnore]` to ignore a specific error.

src/cdsetool/download.py
```diff
@@ -26,31 +30,53 @@
...
 
-    filename = title + ".zip"
-    result_path = os.path.join(path, filename)
+    elif os.path.basename(manifest_file) == "manifest.xml":
+        namespaces = {"ns": "http://www.eumetsat.int/sip"}
+        data_section_elem = xmldoc.find("ns:dataSection", namespaces)
+        for elem in data_section_elem.iterfind("ns:dataObject", namespaces):  # type: ignore
+            path = Path(elem.find("ns:path", namespaces).text)  # type: ignore
+            # Remove prefix present for some files in S3 manifests
+            path = Path(*path.parts[-1:])
```
pjonsson, line 64: If the removal should be here, then put the removal in the call to append: `paths.append(Path(*path...))`.

The previous iteration of this code had a leading `./`, but `Path` takes care of that, what prefix is the comment referring to now?

src/cdsetool/download.py
```diff
@@ -26,31 +30,53 @@
...
-    title = feature.get("properties").get("title")
-    temp_dir_usr = _get_temp_dir(options)
+    paths = []
+    xmldoc = etree.parse(manifest_file)
 
-    if not url or not title:
-        log.debug(f"Bad URL ('{url}') or title ('{title}')")
-        return None
+    if os.path.basename(manifest_file) == "manifest.safe":
+        data_obj_section_elem = xmldoc.find("dataObjectSection")
```
pjonsson, line 53: This looks like a single-use variable, does `for elem in xmldoc.find("dataObjectSection").iterfind("dataObject"):` get reformatted to something spanning more than 2 lines, or could the variable just as well be inlined into where it is used? If it is just a few characters too long, the `xmldoc` variable name could be shortened to `xml`.

As a meta-comment, even if every line looks fine by itself, there are an awful lot of lines and variables so I have to scroll up and down to figure out if things are single-use or used multiple times, and to remember what things were defined as.

src/cdsetool/download.py
```diff
@@ -26,31 +30,53 @@
...
+        for elem in data_obj_section_elem.iterfind("dataObject"):  # type: ignore
+            path = Path(elem.find("byteStream/fileLocation").attrib["href"])  # type: ignore
```
pjonsson, line 55: Can this line start with `paths.append(` instead of `path = `, or does that cause Ruff to break the thing up into multiple lines?

src/cdsetool/download.py
```diff
@@ -96,13 +115,111 @@ def download_feature(
...
+        if not url or not title:
+            log.debug(f"Bad URL ('{url}') or title ('{title}')")
+            return None
+
+        filename = title + ".zip"
+        result_path = os.path.join(path, filename)
+
+        if not options.get("overwrite_existing", False) and os.path.exists(result_path):
+            log.debug(f"File {result_path} already exists, skipping..")
+            return filename
```
pjonsson, line 213: Define filename as `filename = title + ("" if filter_pattern else ".zip")` and put this block before creating the temporary directory (and remove the corresponding block from inside the filter-pattern if-branch).

src/cdsetool/download.py
```diff
@@ -96,13 +115,111 @@ def download_feature(
...
+            try:
+                filtered_files = filter_files(manifest_file, options["filter_pattern"])
+            except Exception as e:  # pylint: disable=broad-exception-caught
+                log.error(f"Failed to filter files in {title}: {e}")
+                return None
+
+            for filtered_file in filtered_files:
+                output_file = os.path.join(
+                    temp_product_path,
+                    filtered_file,
```
pjonsson, line 184: Remove the trailing comma and Ruff should format this into a single line.

src/cdsetool/download.py
```diff
@@ -96,13 +115,111 @@ def download_feature(
...
+            shutil.move(temp_product_path, path)
+
+            return title
+
+        url = _get_feature_url(feature)
+
+        if not url or not title:
+            log.debug(f"Bad URL ('{url}') or title ('{title}')")
+            return None
+
```
pjonsson, line 207: Put this before the temporary directory as well, and define url as `url = _get_odata_url(..) if filter_prefixes else _get_feature_url(feature)`.

src/cdsetool/download.py
```diff
@@ -96,13 +115,111 @@ def download_feature(
...
+                manifest_filename = MANIFEST_FILENAMES[
+                    feature["properties"]["collection"]
+                ]
+            except KeyError:
+                log.error(
+                    "Downloading specific files within product bundle with node "
+                    f"filtering is not supported for this collection type: {title}"
+                )
+                return None
+            manifest_file = os.path.join(temp_product_path, manifest_filename)
```
pjonsson, line 164: Put the `os.path.join(temp_product_path, ` to the left of `MANIFEST_FILENAME[` inside the try-block instead.

tests/download/download_test.py
```diff
@@ -0,0 +1,232 @@
...
+    _get_odata_url,
+    download_feature,
+    download_file,
+    filter_files,
+)
+
+from ..mock_auth import mock_jwks, mock_openid, mock_token
+
+
+def test_get_odata_url():
```
pjonsson, line 19: All the new test functions seem to be missing type annotations.

tests/download/download_test.py
```diff
@@ -0,0 +1,232 @@
...
+    )
+
+    mock_file = str(tmp_path / "mock_file")
+
+    result = download_file(mock_url, mock_file, {})
+
+    # Check that file was written correctly
+    with open(mock_file, "rb") as f:
+        file_content = f.read()
+    expected_content = b"data" * 5
```
pjonsson, line 115: Rename this to `content`, define it before calling `requests_mock.get()`, and pass content to `get()`.

src/cdsetool/download.py
```diff
@@ -26,31 +30,53 @@
...
+def filter_files(manifest_file: str, pattern: str, exclude: bool = False) -> List[str]:
     """
-    Download a single feature
+    Filter a product's files, listed in its manifest, based on a given pattern.
 
-    Returns the feature ID
+    Returns a list of files paths within the product bundle.
+
+    If "exclude" is set to False, only files that match pattern are returned. If it's
+    set to True, only files that do not match pattern are returned.
```
pjonsson, line 47: "All files not matching the pattern are returned if `exclude` is set to true."

src/cdsetool/download.py
```diff
@@ -26,31 +30,53 @@
...
-    feature, path: str, options: Union[Dict[str, Any], None] = None
-) -> Union[str, None]:
+
+def filter_files(manifest_file: str, pattern: str, exclude: bool = False) -> List[str]:
     """
-    Download a single feature
+    Filter a product's files, listed in its manifest, based on a given pattern.
 
-    Returns the feature ID
+    Returns a list of files paths within the product bundle.
```
pjonsson, line 44: "file paths"

src/cdsetool/cli.py
```diff
@@ -69,7 +69,7 @@ def query_search(
 
 # TODO: implement limit
 @app.command("download")
-def download(
+def download(  # pylint: disable=[too-many-arguments, too-many-positional-arguments]
```
pjonsson, line 72: This is fine, it's just PyLint being annoying.

tests/download/download_test.py
```diff
@@ -0,0 +1,232 @@
...
+    mock_token(requests_mock)
+    mock_jwks(mocker)
+
+    # Mock download request
+    mock_url = "http://example.com/file"
+    mocker.patch("cdsetool.download._follow_redirect", return_value=mock_url)
+    requests_mock.get(
+        mock_url,
+        status_code=404,
+        headers={"Content-Length": "100"},
```
pjonsson, line 133: Do the arguments collapse to a single line if you remove the trailing comma?

tests/download/download_test.py
```diff
@@ -0,0 +1,232 @@
...
+    )
+
+    mock_file = str(tmp_path / "mock_file")
+
+    # Avoid retry delay
+    mocker.patch("time.sleep", return_value=None)
+
+    result = download_file(mock_url, mock_file, {})
+
+    assert result is None
```
pjonsson, line 143: The code is straightforward, can remove the additional blank lines for cases like this.

tests/download/download_test.py
```diff
@@ -0,0 +1,232 @@
...
+        """Mock the download_file function to create mock files."""
+        with open(path, "wb") as f:
+            f.write(b"dummy data")
+        return path
+
+    options = {"filter_pattern": "*.jp2"}
+    mock_feature = {
+        "id": "a6215824-704b-46d7-a2ec-efea4e468668",
+        "properties": {
+            "title": "S2B_MSIL1C_20241209T162609_N0511_R040_T17UPV_20241209T195414.SAFE",
```
pjonsson, line 157: Define a variable `title = "S2B_..."` and use that instead of the long string. That should also collapse the multi-line calls to os.path.join below if you remove the trailing comma on the last parameter.

tests/download/download_test.py
```diff
@@ -0,0 +1,232 @@
...
+        return_value=["./GRANULE/file1.jp2", "./GRANULE/file2.jp2"],
+    )
+    mocker.patch(
+        "cdsetool.download.download_file",
+        side_effect=lambda url, path, options: None,
+    )
+
+    final_dir = tmp_path / "test_download_feature_with_filter_failure"
+    product_name = download_feature(mock_feature, final_dir, options)
+    assert os.listdir(tmp_path) == []
```
pjonsson, line 211: IO operations are often slow and I don't think future PRs are likely to break the property tested here, so I think this assert can be omitted.

Same comment for the listdir in the next test.

tests/download/download_test.py
```diff
@@ -0,0 +1,232 @@
...
+
+
+def test_filter_files_no_match():
+    manifest_file_path = "tests/download/mock/sentinel_2/manifest.safe"
+    filtered_files = filter_files(manifest_file_path, "Oa1*.nc")
+    assert not filtered_files
+
+
+def test_download_file_success(requests_mock: Any, mocker: Any, tmp_path: Path):
+    # Mock authentication
```
pjonsson, line 93: I think this comment can be omitted, it's obvious enough that the next 3 lines are mocking authentication.

Same comment for the next test.

src/cdsetool/download.py
```diff
@@ -67,27 +93,20 @@ def download_feature(
...
                 continue
             url = _follow_redirect(url, session)
-            name_dir_prefix = filename.replace(".zip", "____")
-            with (
-                session.get(url, stream=True) as response,
-                tempfile.TemporaryDirectory(
-                    prefix=name_dir_prefix, dir=temp_dir_usr
-                ) as temp_dir,
-            ):
+            with session.get(url, stream=True) as response:
```
pjonsson, line 96: The temporary directory stuff is intricate and I have to run now, so I need to look at the consequences of removing that later.

pjonsson:
>       * Pylint `too-many-return-statements` in `download_feature`: These return statements come from handling errors by logging a message and returning None. Maybe they can be reduced by raising exceptions instead. But I propose to ignore this warning for now.

I think this is a consequence of (remaining) code duplication, so let's leave the ignore in for now and try to reduce the code duplication first.

src/cdsetool/download.py
```diff
@@ -96,13 +115,111 @@ def download_feature(
...
+                    output_file,
+                    options,
+                )
+                if output_file is None:
+                    log.error(f"Failed to download all selected files in {title}")
+                    return None
+
+            # Move downloaded files to output dir
+            os.makedirs(path, exist_ok=True)
+            shutil.move(temp_product_path, path)
```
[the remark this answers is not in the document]
pjonsson: Ignore this comment, I lost track of which if-branch I was in.

tvoirand:
@pjonsson thanks for the prompt review! I took the liberty to mark older discussions as resolved.

src/cdsetool/download.py
tvoirand: Done with 69d2ce7

src/cdsetool/download.py
tvoirand: Done with 463e8cd

src/cdsetool/download.py
tvoirand: Done with cb69586

src/cdsetool/download.py
tvoirand: Done with 851b91d

src/cdsetool/download.py
tvoirand: Are you suggesting to write this?
```
try:
  manifest_file = os.path.join(
    temp_product_path,
    MANIFEST_FILENAMES[feature["properties"]["collection"]],
  )
```
If so, this would eliminate the `manifest_filename` variable, which is still needed by `_get_odata_url`.

tests/download/download_test.py
tvoirand: Done with feb0c66

tests/download/download_test.py
tvoirand: Done with 6c3664c

src/cdsetool/download.py
tvoirand: Done with 091cbb6

src/cdsetool/download.py
tvoirand: Done with 091cbb6

tests/download/download_test.py
tvoirand: Done with ffe8b14

tests/download/download_test.py
tvoirand: Done with ffe8b14

tests/download/download_test.py
tvoirand: Done with ffe8b14

tests/download/download_test.py
tvoirand: Done with ffe8b14

tests/download/download_test.py
tvoirand: Done with 3f5dc6b

src/cdsetool/download.py
tvoirand: If we remove the variable as you propose, it works fine for the first case (`manifest.safe`), but in the second case (`manifest.xml`), it gets reformatted.

Would you prefer:
```python
for elem in xmldoc.find("ns:dataSection", namespaces).iterfind(
    "ns:dataObject", namespaces
):
```
or keeping:
```python
data_section_elem = xmldoc.find("ns:dataSection", namespaces)
for elem in data_section_elem.iterfind("ns:dataObject", namespaces):
```
It also works fine when renaming namespaces to ns:
```python
ns = {"ns": "http://www.eumetsat.int/sip"}
for elem in xmldoc.find("ns:dataSection", ns).iterfind("ns:dataObject", ns):
```

Would be happy to hear your thoughts!

tests/download/download_test.py
pjonsson: Return types too, please (all `test_` functions return None).

src/cdsetool/download.py
pjonsson: You're right, that was a bad suggestion.

src/cdsetool/download.py
pjonsson: See if you can flip the logic into something resembling this sketch that always downloads one file, and then returns early if it's the complete product that was downloaded:
```python
everything = "filter_pattern" in options
filename = title + ".zip" if everything else manifest_file
url = _get_feature_url(..) if everything else get_odata_url(..., filename)
...
with tempdir.TemporaryDirectory():
    if not download_file(url, filename, options):
        return None
    if everything:
        shutil.copy(...)
        return filename
    files = filter_files(..)
    for file in files:
        if not download_file(..):
            return None
````

File: src/cdsetool/download.py (hunk 14 above)
Comment (tvoirand) [reply src/cdsetool/download.py:54]: Done with 0b39e49.

File: src/cdsetool/download.py (hunk 15 above)
Comment (tvoirand) [reply src/cdsetool/download.py]: Done with 261eb49.

File: src/cdsetool/download.py (hunk 16 above)
Comment (tvoirand) [reply src/cdsetool/download.py]: Right ! I got mixed up here. Simplified by using `path.name` with 0b39e49.

File: src/cdsetool/download.py (hunk 17 above)
Comment (tvoirand) [reply src/cdsetool/download.py]: Done with 0b39e49.

File: src/cdsetool/download.py (hunk 18 above)
Comment (tvoirand) [reply src/cdsetool/download.py]: Done with 0b39e49.

File: tests/download/download_test.py (hunk 23 above)
Comment (tvoirand) [reply tests/download/download_test.py]: Right ! Done with 88ad29f.

File: src/cdsetool/download.py (hunk 21 above)
Comment (tvoirand) [reply src/cdsetool/download.py:181]: I think I see the benefit of flipping the logic here, as you put it, as it allows sharing more code blocks between the complete product case and the filtered files case.

I gave it an attempt, but I find the result much less clear than before. Therefore I didn't feel comfortable pushing it to this branch. If that's not too much trouble I am submitting it for your review below, and in this [other branch available on my fork](https://github.com/tvoirand/CDSETool/commit/e4a07f0ab2540c2d25a51c9ea9fd7ae876d68a51).

The points bugging me are:

- There are differences in the download of the first file that cause the new logic to require many new if-statements (5 in total, linked with `manifest_filename`, `filename`, `result_path`, `url` and the `return` statements being different in both cases). This is quite confusing to read (in my opinion) since broadly different values are hidden under these double-logic variables.

- Manifest filename reading requires initializing with an empty string in addition to being wrapped in an if-statement, or else more errors are raised by linters (pylint `possibly-used-before-assignment` and pyright `reportPossiblyUnboundVariable`). The if-statements should make this fine, but these warnings rightfully indicate a "code smell" in my opinion.

I am not sure if the change is worth applying. It comprises roughly as many additions as deletions (+51/-52). But I would be happy to hear your opinion again!

```
def download_feature(  # pylint: disable=too-many-return-statements
    feature, path: str, options: Union[Dict[str, Any], None] = None
) -> Union[str, None]:
    """
    Download a single feature.

    Returns the feature title.
    """
    options = options or {}
    log = _get_logger(options)
    temp_dir_usr = _get_temp_dir(options)
    title = feature.get("properties").get("title")
    download_full = "filter_pattern" not in options

    manifest_filename = ""
    if not download_full:
        try:
            manifest_filename = MANIFEST_FILENAMES[feature["properties"]["collection"]]
        except KeyError:
            log.error(
                "Downloading specific files within product bundle with node "
                f"filtering is not supported for this collection type: {title}"
            )
            return None

    filename = title + ".zip" if download_full else manifest_filename
    result_path = (
        os.path.join(path, filename) if download_full else os.path.join(path, title)
    )
    if not options.get("overwrite_existing", False) and os.path.exists(result_path):
        log.debug(f"File {result_path} already exists, skipping..")
        return os.path.basename(result_path)

    url = (
        _get_feature_url(feature)
        if download_full
        else _get_odata_url(feature["id"], title, manifest_filename)
    )

    if not url or not title:
        log.debug(f"Bad URL ('{url}') or title ('{title}')")
        return None

    with tempfile.TemporaryDirectory(
        prefix=f"{title}____", dir=temp_dir_usr
    ) as temp_dir:
        temp_file = os.path.join(temp_dir, filename)
        if not download_file(url, temp_file, options):
            return None
        if download_full:
            shutil.copy(temp_file, result_path)
            return filename

        # Create product dir structure to download filtered contents from OData API
        temp_product_path = os.path.join(temp_dir, title)
        os.makedirs(temp_product_path, exist_ok=True)

        # List files that match pattern based on manifest file contents
        filtered_files = filter_files(temp_file, options["filter_pattern"])

        for filtered_file in filtered_files:
            output_file = os.path.join(temp_product_path, filtered_file)
            os.makedirs(os.path.dirname(output_file), exist_ok=True)
            result = download_file(
                _get_odata_url(feature["id"], title, filtered_file),
                output_file,
                options,
            )
            if not result:
                log.error(f"Failed to download all selected files in {title}")
                return None

        # Move downloaded files to output dir
        os.makedirs(path, exist_ok=True)
        shutil.move(temp_product_path, path)

        return title
```

Comment (pjonsson) [reply src/cdsetool/download.py:181]: >     * There are differences in the download of the first file that cause the new logic to require many new if-statements (5 in total, linked with `manifest_filename`, `filename`, `result_path`, `url` and the `return` statements being different in both cases). This is quite confusing to read (in my opinion) since broadly different values are hidden under these double-logic variables.

I agree that there are more if-statements, but we're trading two copies of the same code against something that looks like this:
```python
variable1 = x1 if cond else y1
variable2 = os.path.join(something, x2 if cond else y2)
variable3 = ...
with .. as temp_dir:
    if not download_file(url, filename):
        return None
    if everything:
        shutil.copy(...)
        return filename
    for file in filter_files(...):
        if not download_file(..., file):
            return None
```
and while the definition of variable1..variable3 are more complicated than the original code, each line can still be verified in isolation. Once we get those lines right, we are unlikely to need to change them again. (And the url variable always contains the url to use, just like the filename variable always contains the filename to use, it's just a different url/filename for the two cases.)

The code inside the with-statement is the intricate part that is likely to change when we discover a new and fascinating way that the remote servers misbehave, but that code is supposed to behave the same way regardless of the content of the url or filename variables. Sharing the code for downloading the first file means it's unlikely we'll make a code change that can still download the manifest but not the entire archive, or vice versa.

I agree that uninitialized variables is a code smell, but I think things will buff out if we push the if-statements even further into the leaves in the parse-tree. And I don't claim to have the foresight of being able to tell exactly what the code will look like after the changes, but I'm pretty sure my "method" will yield something manageable.

I commented on the file in your separate branch, but the comment interface looked unfamiliar, I hope you are able to see my comments.

tvoirand: Thank you again for your review.

Based on your approval of the new logic in `download_feature`, I am pushing the code to this branch.

I have taken into account most of your latest comments with e34ca32, with two points remaining:

> ```python
> os.makedirs(temp_product_path, exist_ok=True)
> ```
> Isn't this handled already by the makedirs inside the loop?

Are you referring to the temporary directory creation in the following line `with tempfile.TemporaryDirectory(prefix=f"{title}____", dir=temp_dir_usr) as temp_dir:` ?

This line creates a directory with a randomly generated name that includes the product title and trailing underscores. However, to correctly structure the downloaded files, a directory named only after the product title is required. This is created by: `os.makedirs(temp_product_path, exist_ok=True)`. This step is not handled elsewhere in the code.

> ```python
> shutil.move(temp_product_path, path)
> ```
> Moving a directory across file systems doesn't work.

I am reading in [the doc](https://docs.python.org/3/library/shutil.html#shutil.move) that `shutil.move` should work across filesystems, where it copies files to the destination and removes them from the source.

tvoirand: Looking at it again, I see the `makedirs` you are referring too. You can ignore my previous answer to the first point.

You are right that this is handled by the `makedirs` inside the loop. However removing this line will cause `shutil.move` to fail in the case where no file is returned by `filter_pattern`, because `temp_product_path` is never created. 

But maybe we could use this case to raise a warning that the filter didn't match any file.

src/cdsetool/download.py
```diff
@@ -96,11 +102,82 @@ def download_feature(
...
+        os.makedirs(temp_product_path, exist_ok=True)
+
+        # List files that match pattern based on manifest file contents
+        for file in filter_files(temp_file, options["filter_pattern"]):
+            output_file = os.path.join(temp_product_path, file)
+            os.makedirs(os.path.dirname(output_file), exist_ok=True)
+            if not download_file(
+                _get_odata_url(feature["id"], title, file),
+                output_file,
+                options,
```
pjonsson, line 171: Do these arguments format into a single line if you removing the trailing comma?

src/cdsetool/download.py
```diff
@@ -96,11 +102,82 @@ def download_feature(
...
+        log.error(
+            f"No support for downloading individual files in {collection} products"
+        )
+        return None
+
+    filename = title + ".zip" if download_full else manifest_filename
+    result_path = os.path.join(path, filename if download_full else title)
+    if not options.get("overwrite_existing", False) and os.path.exists(result_path):
+        log.debug(f"File {result_path} already exists, skipping..")
+        return os.path.basename(result_path)
```
pjonsson, line 138: The previous code had this check after the URL check, so I think we should move this down.

src/cdsetool/download.py
tvoirand: Yes ! Done with f29d600

src/cdsetool/download.py
tvoirand: Done with e30ad1e

src/cdsetool/download.py
tvoirand: Understood about the warning.

I have applied the change you propose to use the same return-logic with 42003ea, thanks for the suggestion.

tests/download/download_test.py
```diff
@@ -0,0 +1,248 @@
...
+
+
+def test_get_odata_url() -> None:
+    product_id = "a6215824-704b-46d7-a2ec-efea4e468668"
+    product_name = "S2B_MSIL1C_20241209T162609_N0511_R040_T17UPV_20241209T195414.SAFE"
+    href = "path/to/resource.xml"
+    expected_url = (
+        "https://download.dataspace.copernicus.eu/odata/v1/"
+        "Products(a6215824-704b-46d7-a2ec-efea4e468668)/"
+        "Nodes(S2B_MSIL1C_20241209T162609_N0511_R040_T17UPV_20241209T195414.SAFE)/"
```
pjonsson, line 34: `f"Nodes(product_name)/"

tests/download/download_test.py
```diff
@@ -0,0 +1,248 @@
...
+    mock_jwks(mocker)
+
+    mock_url = "http://example.com/file"
+    mocker.patch("cdsetool.download._follow_redirect", return_value=mock_url)
+    requests_mock.get(mock_url, status_code=404, headers={"Content-Length": "100"})
+    mock_file = str(tmp_path / "mock_file")
+    mocker.patch("time.sleep", return_value=None)  # Avoid retry delay
+
+    result = download_file(mock_url, mock_file, {})
+    assert not result
```
pjonsson, line 137: `assert result is False`

tests/download/download_test.py
```diff
@@ -0,0 +1,248 @@
...
+            "collection": "SENTINEL-2",
+        },
+    }
+    mocker.patch(
+        "cdsetool.download.filter_files",
+        return_value=["./GRANULE/file1.jp2", "./GRANULE/file2.jp2"],
+    )
+    mocker.patch(
+        "cdsetool.download.download_file",
+        mock_download_file,
```
pjonsson, line 197: Remove the trailing comma and see if this formats shorter.

tests/download/download_test.py
```diff
@@ -0,0 +1,248 @@
...
+
+
+def test_download_feature_with_filter(mocker: Any, tmp_path: Path) -> None:
+    options = {"filter_pattern": "*.jp2"}
+    title = "S2B_MSIL1C_20241209T162609_N0511_R040_T17UPV_20241209T195414.SAFE"
+    mock_feature = {
+        "id": "a6215824-704b-46d7-a2ec-efea4e468668",
+        "properties": {
+            "title": title,
+            "collection": "SENTINEL-2",
```
pjonsson, line 188: Remove the trailing comma and see if this formats shorter.

tests/download/download_test.py
```diff
@@ -0,0 +1,248 @@
...
+        },
+    }
+    mocker.patch(
+        "cdsetool.download.filter_files",
+        return_value=["./GRANULE/file1.jp2", "./GRANULE/file2.jp2"],
+    )
+    mocker.patch(
+        "cdsetool.download.download_file",
+        side_effect=lambda url, path, options: None,
```
pjonsson, line 222: Remove the trailing comma and see if this formats better.

tests/download/download_test.py
```diff
@@ -0,0 +1,248 @@
...
+
+def test_download_feature_with_filter_unsupported_coll(
+    caplog: Any, tmp_path: Path
+) -> None:
+    options = {"logger": logging.getLogger(__name__), "filter_pattern": "*MTL.txt"}
+    mock_feature = {
+        "id": "a6215824-704b-46d7-a2ec-efea4e468668",
+        "properties": {
+            "title": "L8XXX",
+            "collection": "Lansat8",
```
pjonsson, line 238: Missing `d` in the collection name, and try to remove the trailing comma and see if this formats better.

src/cdsetool/download.py
```diff
@@ -26,31 +30,46 @@
...
+    "SENTINEL-2": "manifest.safe",
+    "SENTINEL-3": "manifest.xml",
+}
 
-def download_feature(
-    feature, path: str, options: Union[Dict[str, Any], None] = None
-) -> Union[str, None]:
+
+def filter_files(
+    manifest_file: str, pattern: Union[str, None], exclude: bool = False
```
pjonsson, line 41: This function is new so we don't have to worry about API compatibility. There's only a single caller, and I think that caller should pass a `Path` for manifest_file, and this function should return `List[Path]`. That will replace the `os.path.basename(..)` in this function with `manifest_file.name` (or whatever the property is called), and avoid the conversion to str in the return statement.

src/cdsetool/download.py
```diff
@@ -26,31 +30,46 @@
...
-        return None
+    paths = []
+    xmldoc = ET.parse(manifest_file)
+    if os.path.basename(manifest_file) == "manifest.safe":
+        for elem in xmldoc.find("dataObjectSection").iterfind("dataObject"):  # pyright:ignore[reportOptionalMemberAccess]
+            paths.append(Path(elem.find("byteStream/fileLocation").attrib["href"]))  # pyright:ignore[reportOptionalMemberAccess]
+    elif os.path.basename(manifest_file) == "manifest.xml":
+        ns = {"ns": "http://www.eumetsat.int/sip"}
+        for elem in xmldoc.find("ns:dataSection", ns).iterfind("ns:dataObject", ns):  # pyright:ignore[reportOptionalMemberAccess]
+            paths.append(Path(elem.find("ns:path", ns).text).name)  # pyright:ignore[reportOptionalMemberAccess,reportArgumentType]
```
pjonsson, line 62: Isn't the `reportArgumentType` because this line is adding a non-Path to a list of Paths? Wrap it in a `Path()` instead.

src/cdsetool/download.py
```diff
@@ -26,31 +30,46 @@
...
 
-    filename = title + ".zip"
-    result_path = os.path.join(path, filename)
+    return [str(path) for path in paths if fnmatch.fnmatch(path, pattern) ^ exclude]
 
-    if not options.get("overwrite_existing", False) and os.path.exists(result_path):
-        log.debug(f"File {result_path} already exists, skipping..")
-        return filename
+
+def download_file(url: str, path: str, options: Dict[str, Any]) -> bool:
```
pjonsson, line 67: Make `path` have a `Path` type here too.

src/cdsetool/download.py
pjonsson: Nice!

I think I remembered why we couldn't use move before: the original code used `with tmpfile.TemporaryFile() ..` instead of `TemporaryDirectory()`, so that explains why I thought we couldn't move files.

Thanks for hanging in there with all the changes to this piece of code, I just had a look from top to bottom of the diff and commented on a bunch of smaller things, but I feel we're getting close to the finish line.

tests/download/download_test.py
tvoirand: Done with fb78ac6

tests/download/download_test.py
```diff
@@ -0,0 +1,248 @@
...
+    mock_file = str(tmp_path / "mock_file")
+
+    result = download_file(mock_url, mock_file, {})
+
+    # Check that file was written correctly
+    with open(mock_file, "rb") as f:
+        file_content = f.read()
+    assert file_content == content
+
+    assert result
```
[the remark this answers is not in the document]
tvoirand: Done with 971a722

tests/download/download_test.py
tvoirand: Done with 971a722

tests/download/download_test.py
tvoirand: Done with 5d4f0cd

tests/download/download_test.py
tvoirand: Done with 5d4f0cd

tests/download/download_test.py
```diff
@@ -0,0 +1,248 @@
...
+    mocker.patch(
+        "cdsetool.download.download_file",
+        mock_download_file,
+    )
+
+    final_dir = str(tmp_path / "test_download_feature_with_filter")
+    product_name = download_feature(mock_feature, final_dir, options)
+    assert os.path.exists(os.path.join(final_dir, title, "GRANULE", "file1.jp2"))
+    assert os.path.exists(os.path.join(final_dir, title, "GRANULE", "file2.jp2"))
+    assert product_name == mock_feature["properties"]["title"]
```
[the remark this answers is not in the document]
tvoirand: Done with 971a722

tests/download/download_test.py
tvoirand: Done with 5d4f0cd

src/cdsetool/download.py
tvoirand: I think this comment (and the other two below) is related with a previous discussion about using `pathlib` in `filter_files` (started on Jan. 6, with your comment "_Why does ./ need to be removed, isn't that just the current directory?_"). 

Making `filter_files` and `download_file` take and return `Path` instead of `str` would introduce a mix of `str` and `Path` types within `download_feature`. For example, the `path` argument passed to `download_feature` would remain a `str`, while the `path` argument passed to `download_file` would be a `Path`. Don't you think this inconsistency could cause confusion and should be avoided?

That said, I see the value in using `Path` consistently across the codebase. However this might deserve a dedicated PR (to discuss API compatibility for example, as you mentioned above) to avoid overloading this one, which already generated many comments.

src/cdsetool/download.py
tvoirand: The `reportArgumentType` is caused by `elem.find("ns:path", ns).text` being possibly `None` (if something's wrong in the manifest contents), which is not accepted for the `Path` constructor.

The error is:
```
error: Argument of type "str | None" cannot be assigned to parameter "args" of type "StrPath" in function "__new__"
    Type "str | None" is not assignable to type "StrPath"
      Type "None" is not assignable to type "StrPath"
        "None" is not assignable to "str"
        "None" is incompatible with protocol "PathLike[str]"
          "__fspath__" is not present (reportArgumentType)
```

src/cdsetool/download.py
tvoirand: Happy to hear that we're getting close! Thanks for taking the time to review.

src/cdsetool/download.py
pjonsson: Your last paragraph touches much larger questions. The first question is whether it would be a good idea for the entire code base to use `Path`, and I genuinely don't know the answer to that, there might be places where a string would be better. The second "question" is practicality, so even if the code base should use `Path` from a purely technical standpoint, it wasn't written that way from the start and now we're stuck with a bunch of users outside the repository that expect the "path"  they pass to `download_feature` to be a string.

When it comes to `filter_files` and `download_file` there are currently no callers outside this repository (and they're not really functions called by the end-user), so the question before this PR is merged is mainly about how we want things to be. My experience is that it's better to have 50% of the code base with a good API and 50% with a bad API than it is to have 100% of the code base with a bad API, because those extra 50% of bad API cements the bad API even more.

src/cdsetool/download.py
pjonsson: Broken manifest files are unfortunately not completely out of character, so it's probably best to add an explicit check:
```python
path = elem.find("ns:path", ns).text
if path is not None:
    paths.append(Path(path).name)
```

tvoirand: The other pyright ignores (`reportOptionalMemberAccess`) are linked with cases where the manifest is broken and `xmldoc.find()` or `elem.find()` return `None`. The associated pyright error messages are `"iterfind|attrib|text is not an attribute of None"`. Would it make sense to add explicit checks for each `.find()` call in this case?

I initially thought that leaving these potential errors unhandled made sense because it causes `filter_files` to stop functioning and report an error when reading a broken manifest. 

Adding explicit checks allows `filter_files` to continue even if part of the manifest is broken, which is useful if the broken part isn't relevant to the user's files of interest. On the other hand, there's a (small) risk that the broken part does concern the user's files of interest, which could lead to silent failures (for example a file that should be available is skipped due to a broken `ns:path` entry, and this failure isn't explicitly reported).

We could also add a warning when `.find()` returns `None` instead of silently skipping such cases:
```python
path = elem.find("ns:path", ns).text
if path is None:
    logger.warning(f"Missing 'ns:path' in {manifest_file}: {elem}")
    continue
paths.append(Path(path).name)
```

src/cdsetool/download.py
tvoirand: Understood, I made the change with 048128e

pjonsson:
I didn't get time to look at this after the most recent round of changes, sorry about that. Will try to squeeze it in next week.

tvoirand:
Thanks for your message, no problem at all, please take your time !

src/cdsetool/download.py
tvoirand: After taking another look at this point:

Adding checks for each `find()` call would add many nested if-statements, which doesn't look right.

Adding a warning within `filter_files` would involve passing `options` instead of just `pattern`, which makes this function slightly more complex.

I liked the idea of `filter_files` failing on broken manifests, which seems to be the current behavior. 

If we want to handle this more properly, we could catch exceptions within `download_feature`. (I previously suggested this with a broad `except Exception:`, but we could narrow it down)
I think possible errors are `AttributeError` (`find()` returning `None`), `KeyError` (`href` being absent), `TypeError` (`path` element's `text` being `None`). This however, would add another return statement and involve disabling again pylint's too-many-return-statements in `download_feature`. It would look like:
```python
try:
    filtered_files = filter_files(
        Path(temp_file), options.get("filter_pattern")
    )
except (KeyError, AttributeError, TypeError) as e:
    log.error(f"Failed to parse manifest file of {title}: {e}")
    return None
```

But of course I would be happy to hear your thoughts, and would be fine with other approaches, including your initial suggestion to explicitly check only S3 manifests' `path`.

pjonsson: I had another look at the code, and I think you have some great points. If you can factor out this part of the code to a helper function:
```python
        for elem in xmldoc.find("dataObjectSection").iterfind("dataObject"):  # pyright:ignore[reportOptionalMemberAccess]
            paths.append(Path(elem.find("byteStream/fileLocation").attrib["href"])) 
```
that could be called from the two branches in `filter_files`, I think it will be manageable to check for None as return values and return accordingly.

I don't know the XML API by heart, and the second branch of `filter_files` looks slightly different from the first branch. Could the first branch's `.attrib["href"]` be used for  the second branch's `.text`as well, so `.attrib["text"]` would work, or are those separate parts of the API? If `.attrib[string]` could work, I think the helper function could take the xmldoc + the 4 strings as parameters, and return a list of Paths.

The second branch also takes `.name` to remove the directory, and I think that can be done in the second branch on the return value from the helper function.

I don't have an immediate answer on whether errors should be logged or hard failures and I need to do some other stuff now, but if it's possible to make something like the helper function I described work, I'm fairly sure we can get either one of your suggestions for handling errors to work quite easily after that.

tvoirand: I investigated factoring the part you pointed out into a single helper function. But due to the differences in how file locations are stored in SAFE and S3 manifest, this doesn't seem feasible.

In SAFE manifests, file locations are stored as XML attributes (named "href"), and in S3 manifests they are stored as XML elements (named "path"). Because of this difference in structure, the parsing methods are different: `.attrib["href"]` for SAFE and `.text` for S3.

Still, I tried moving this part in two separate helper functions, which looks like this:

<details>

<summary>Python code showing helper functions to parse manifest files</summary>

```python
def filter_files(
    manifest_file: Path, pattern: Union[str, None], exclude: bool = False
) -> List[Path]:
    """
    Filter a product's files, listed in its manifest, based on a given pattern.

    Returns a list of file paths within the product bundle.

    All files not matching the pattern are returned if "exclude" is set to true.
    """
    if pattern is None:
        return []
    paths = []
    xmldoc = ET.parse(manifest_file)
    if manifest_file.name == "manifest.safe":
        paths.extend(_parse_safe_manifest(xmldoc))
    elif os.path.basename(manifest_file) == "manifest.xml":
        paths.extend(_parse_s3_manifest(xmldoc))
    return [path for path in paths if fnmatch.fnmatch(path, pattern) ^ exclude]


def _parse_safe_manifest(xmldoc: ET.ElementTree) -> List[Path]:
    # pylint: disable=line-too-long
    paths = []
    for elem in xmldoc.find("dataObjectSection").iterfind("dataObject"):  # pyright:ignore[reportOptionalMemberAccess]
        paths.append(Path(elem.find("byteStream/fileLocation").attrib["href"]))  # pyright:ignore[reportOptionalMemberAccess]
    return paths


def _parse_s3_manifest(xmldoc: ET.ElementTree) -> List[Path]:
    # pylint: disable=line-too-long
    paths = []
    ns = {"ns": "http://www.eumetsat.int/sip"}
    for elem in xmldoc.find("ns:dataSection", ns).iterfind("ns:dataObject", ns):  # pyright:ignore[reportOptionalMemberAccess]
        paths.append(Path(Path(elem.find("ns:path", ns).text).name))  # pyright:ignore[reportOptionalMemberAccess,reportArgumentType]
    return paths
```

</details>

However it didn't strike me as improving the ability to log warnings or handle errors. The same drawbacks remain:
- Adding a warning for broken manifests still involves passing `options` to `filter_files` and the helper functions
- Catching exceptions in `download_feature` still involves adding a return statement and disabling again pylint's too-many-return-statements

My take would be to leave the code as it is for now, with its behavior of hard failure on broken manifests, or, if need be, catch exceptions in `download_feature` in order to log a specific message at the cost of disabling pylint's too-many-return-statements, as written in the snippet of my latest comment above. 

But let me know if I'm missing a clearer solution or if I misunderstood your suggestions !

pjonsson: Sorry, busy week. Here's a quick Python 3.12 sketch of the helper function I meant:
```python
def fun(h1: str, h2: str, h3: str, ns : dict[str, str] | None) -> list[Path] | None:
    x1 = xmldoc.find(h1, ns)  # FIXME: xmldoc is defined in global scope
    if x1 is None:
        return None
    paths = []
    for elem in x1.iterfind(h2, ns):
        x3 = elem.find(h3, ns)
        if x3 is None:
            return None
        p3 = Path(x3.attrib["href"] if ns is None else x3.text)
        paths.append(p3 if ns is None else Path(p3.name))
    return paths  # FIXME: add back matching
```
then that can be called with either of:
```python
x = fun("dataObjectSection", "dataObject", "byteStream/fileLocation", None)
x = fun("ns:dataSection", "ns:dataObject","ns:path", {"ns": "http://www.eumetsat.int/sip"})
```
Changing behavior based on namespace being `None` isn't a work of beauty, but we currently have two cases so it works for now and can be generalized when we get a third case.

My sketch is also missing the filtering functionality, takes the xmldoc from the global scope, `hN`/`xN`/`pN` should have better names, and should obviously not be called `fun`.

Writing this, I realize that the variable `p3` should probably be called path, but I'm not going to change that in the code now since that will surely break something.

src/cdsetool/download.py
```diff
@@ -26,31 +30,45 @@
 from cdsetool.monitor import NoopMonitor, StatusMonitor
 from cdsetool.query import FeatureQuery
 
+MANIFEST_FILENAMES = {
+    "SENTINEL-1": "manifest.safe",
+    "SENTINEL-2": "manifest.safe",
+    "SENTINEL-3": "manifest.xml",
```
pjonsson, line 36: Looking at Sentinel-3:

I have some OL_2_WFR___ downloaded from EUMETSAT laying around, and the manifest file in those is called `xfdumanifest.xml`.

I also have some SL_2_LST___ downloaded from Copernicus, and their manifest file is also called `xfdumanifest.xml`.

What Sentinel-3 product did you look at that had `manifest.xml`?

fbalaban:
sorry for jumping in and the noise, just wanted to say a big thank you to you both for this demanding feature and PR. please ping me if you need any assistance. cheers

src/cdsetool/download.py
tvoirand: I had not noticed this, thank you for the catch !

It looks like S3 products from CDSE contain both `manifest.xml` and `xfdumanifest.xml` files (see this [example CDSE OData API Node](https://download.dataspace.copernicus.eu/odata/v1/Products(b815ba6e-b56f-451b-95b5-e11ad7c704e1)/Nodes(S3B_OL_2_WFR____20241217T095305_20241217T095605_20241218T170707_0179_101_079_2160_MAR_O_NT_003.SEN3)/Nodes)).

The good news is that S3 `xfdumanifest.xml` has the same `dataObjectSection` structure as S1 and S2 `manifest.safe`. This simplifies manifest parsing.

I had looked at an `OL_2_WFR___` product from CDSE (more precisely `S3B_OL_2_WFR____20241217T095305_20241217T095605_20241218T170707_0179_101_079_2160_MAR_O_NT_003.SEN3`). While exploring the OData API when starting working on this feature, I stopped at the first S3 manifest I found (which was `manifest.xml`) and worked from that, it would have saved some time if I had considered the `xfdumanifest.xml`. But I guess the effort spent on reading both manifests was not useless, because it helped prepare supporting other products like Landsat.

By the way I replaced the mock `manifest.xml` with `xfdumanifest.xml` in `tests/download/mock/sentinel_3`. The file is not heavy, but there's a lot of lines, I hope this is fine, otherwise we can trim it to use only the `dataObjectSection`.

src/cdsetool/download.py
tvoirand: Thanks for the clarification with this sketch, and my turn to apologize for the delayed answer.

Things can be simplified because it turns out `xfdumanifest.xml` has the same structure as `manifest.safe`, as mentioned in the other thread. But I still think the helper function is useful, to isolate manifest reading which could become messy if we add support for other products later.

I applied your sketch with two tweaks in my latest commits (eb88020, 6586cdf):
- Use `manifest_file` as argument of the helper function and parse the xml from within that function
- Use `.get("href")` instead of `.attrib["href"]`, this avoids potential key error while still returning None if the "href" attribute is absent

I also added a `return None` in `download_feature` when manifest parsing fails. However I had to apply again pylint's too-many-return-statements ignore.

Looking forward to your feedback !

src/cdsetool/download.py
pjonsson: Oh yeah, put the full file there so it's easy to update in case we do need to replace it. Repositories that are several hundred megabytes/many gigabytes are clunky to work with (and slow to clone both for local work and CI tests), but a couple of thousand lines is far from those numbers, and these files aren't frequently updated causing the repository to grow like crazy.

Unfortunately I'm a bit swamped right now so I need a few more days before I can look at the other code change, but I'm glad finding the "right" manifest helped.

pjonsson:
@tvoirand just letting you know that this is still on my radar and I'm slowly working through my pile to get back to this. I've kept thinking "I'll get to this next week" and then something else drops out of the sky and needs to be dealt with immediately, so I apologize for the long silence.

tvoirand:
No problem, thanks for letting me know. I am looking forward to completing this PR, but there's no rush at all. Good look working through your pile!

src/cdsetool/download.py
```diff
@@ -96,11 +116,81 @@ def download_feature(
...
+            ):
+                log.error(f"Failed to download {file} from {title}")
+                return None
+
+        # Move downloaded files to output dir
+        if download_full or filtered_files:
+            os.makedirs(path, exist_ok=True)
+            shutil.move(temp_file if download_full else temp_product_path, path)
+
+        return filename if download_full else title
```
pjonsson, line 193: My PyCharm complains about not having return statements on all paths. I don't see why, so I suspect PyCharm is wrong.

Can this return statement be de-dented one level, or will that break something?

tvoirand: Yes this return statement can be de-dented one level. It returns a string after download is complete and files have been moved from `temp_dir` to final `path`. De-denting one level results in closing the temp dir before returning, which doesn't break anything. I just applied the change with 2e3c7f1. Hope now your PyCharm doesn't complain anymore !

src/cdsetool/download.py
```diff
@@ -26,31 +30,56 @@
...
+    "SENTINEL-3": "xfdumanifest.xml",
+}
 
-def download_feature(
-    feature, path: str, options: Union[Dict[str, Any], None] = None
-) -> Union[str, None]:
+
+def filter_files(
+    manifest_file: Path, pattern: Union[str, None], exclude: bool = False
+) -> List[Path] | None:
```
pjonsson, line 42: `Union[List[Path], None]` in the return type here.

src/cdsetool/download.py
```diff
@@ -26,31 +30,56 @@
...
     """
-    options = options or {}
-    log = _get_logger(options)
-    url = _get_feature_url(feature)
-    title = feature.get("properties").get("title")
-    temp_dir_usr = _get_temp_dir(options)
 
-    if not url or not title:
-        log.debug(f"Bad URL ('{url}') or title ('{title}')")
+    def read_sentinel_manifest(manifest_file: Path) -> List[Path] | None:
```
pjonsson, line 51: `Union[List[Path], None]` in the return type here as well.

pjonsson:
Unfortunately it seems the CI isn't in tip top shape, but that is not your problem.

I think pylint is failing because of the Python 3.10-syntax I commented about.

The failing formatting can be fixed by running `ruff format` with `ruff==0.9.3`. This formats a bit differently than the latest release of Ruff, so you probably need to run a Ruff 0.9.x release for that command. Since you will have Ruff available in your environment, you can also run `ruff check --fix` to fix import sorting.

Not sure why pytest fails, did the tests work locally for you?

pjonsson (approved):
Everything is green and code looks great, thank you!

PR #236, opened 2025-01-30
Update to Ruff 0.9.3

pjonsson:


pjonsson:
@subnet001 not sure why CI is failing, but I'm fairly sure it's not caused by the merged PR. If you have an override so you can merge this, I suggest doing so since this is only formatting changes.

Main being red is an issue, but I'm most likely busy for the next week, so someone else should look into that.

pjonsson:
And my guess for why main is red is that some recent Pyright release got more capable (or got new typeshed definitions) so it detected some pre-existing problem with our code that was present but wasn't flagged before.

subnet001:
That sounds reasonable to me, i'll merge it.

PR #247, opened 2025-03-18
Update cli.py to download only last available processing baseline for Sentinel2 collection.

ffilipponi:
Added CLI argument '--last-baseline' to download only last available Sentinel2 processing baseline (only applicable to 'Sentinel2' collection).
Query result subset is performed by a new function named 'subset_last_baseline', defined in the 'cli.py' script.

### DIFF
--- a/src/cdsetool/cli.py
+++ b/src/cdsetool/cli.py
@@ -27,17 +27,15 @@ def subset_last_baseline(features):
         processingBaseline.append(features[i].get("properties").get("processingBaseline"))
     fi = list(map(lambda x: startDate.index(x), startDate))
     fl = []
-    for d in range(max(fi)+1):
+    for d in set(fi):
         ff = [i for i,x in enumerate(fi) if x==d]
         if len(ff) == 1:
             fl.append(int("".join(map(str, ff))))
         else:
             pBl = [processingBaseline[i] for i in ff]
-            if 99.99 in set(pBl):
-                pBl[pBl.index(99.99)] = 0
-            pBp = pBl.index(max(pBl))
-            pBv = int("".join(str(ff[pBp])))
-            fl.append(pBv)
+            for b in range(0, len(pBl)):
+                if pBl[b] < 9900 and pBl[b] == max(pBl):
+                    fl.append(int("".join(str(ff[b]))))
     sub_features = [features[i] for i in fl]
     return(sub_features)
 


src/cdsetool/cli.py
```diff
@@ -78,6 +97,9 @@ def download(
     overwrite_existing: Annotated[
         bool, typer.Option(help="Overwrite already downloaded files")
     ] = False,
+    last_baseline: Annotated[
```
pjonsson, line 100: Thank you for your contribution.

I can't see any query parameter in the API that corresponds to what we need, so this needs to be implemented through some kind of post-filtering as you've done. I would like to have a slightly more general interface though, `--min-version x` and `--max-version y`, so that the user can choose to download all version 400-499 products by giving the options `--min-version 400 --max-version 500`.

I have two high-level comments about the implementation:
* The implementation should be somewhere inside `query_features` so CDSETool behaves the same way when downloading and querying, and also when CDSETool is used as a library.  I'm guessing the implementation should change this line: https://github.com/CDSETool/CDSETool/blob/b4cc13acc5bc8af57a4058b6d84682eea0d5f880/src/cdsetool/query.py#L103 to only add the features that have the right processing version, and the result count on this line: https://github.com/CDSETool/CDSETool/blob/b4cc13acc5bc8af57a4058b6d84682eea0d5f880/src/cdsetool/query.py#L105 also needs to be adjusted to account for the skipped features.
* I'm not sure if the processing version will be easily available, but if it isn't, the version can be found as part of  the name of the SAFE archive: https://sentiwiki.copernicus.eu/web/s2-products#S2Products-CompactNamingConventionS2-Products-Compact-Naming-Conventiontrue

ffilipponi: Dear @pjonsson,
thanks for considering my proposal.
I do agree that the implementation should be somewhere inside `query_features`. Nevertheless, the 'processingBaseline' parameter is defined as a string value, and the query can be done only by defining a value 'equal' to a specific processing baseline (no possibility to use 'not'). As a consequence, setting minimum and/or maximum processing baseline versions may result in a very big number of single queries to be done.
The solution I propose, which works on the refinement of the query results, is a workaround to the limitations imposed by the definition of 'processingBaseline' parameter as a string value. It should meet the needs of users, which are not to download multiple versions of one satellite acquisition, but the latest version available.
In conclusion, I suggest implementing the proposed solution as the final step of the `query_features` function.

pjonsson: Sorry for being unclear, I completely agree that the functionality should be implemented as asking a more general query and then filtering the results locally.

I'm concerned that the code base will become difficult to maintain if we add a specialized lazy iterator for every new requirement that we discover, so that was why I suggested implementing the filtering somewhere inside the `FeatureQuery` that we already have.

PR #259, opened 2025-07-16
CI: pin actions by hash

pjonsson:
This prevents us from leaking
passwords if the repository
of one of the actions we
use get compromised.


subnet001:
@pjonsson Good work, approved

PR #260, opened 2025-07-19
CI: add security scan workflow

pjonsson:
This will populate the security
tab in Github with scan results.

### DIFF
--- a/.github/workflows/scan.yml
+++ b/.github/workflows/scan.yml
@@ -1,0 +1,35 @@
+name: Scan
+
+on:
+  schedule:
+    - cron: '0 0 * * *'
+  workflow_dispatch:
+
+permissions: {}
+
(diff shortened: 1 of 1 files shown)


pjonsson:
I doubt we'll hear much from this, so the scanner is really low effort to have around, but it fixes one thing on the OpenSSF Scorecard report (example report for other repository: https://securityscorecards.dev/viewer/?uri=github.com/opendatacube/datacube-core).

I don't want their Github Action so we get a constant "3" on our security tab warning about stuff that isn't relevant for this repository, but the stuff they recommend seems both reasonable and balanced.

CDSETool apparently isn't in the top 1 million most popular open source projects, so there isn't a scorecard report on that website, but one can run their CLI if curious about the scoring for this repository.

PR #261, opened 2025-07-19
CI: drop permissions for workflows

pjonsson:
The read contents permission is enabled
by default nowadays for repositories,
so use the empty dictionary to indicate
that.

Remove the write permissions for
the pytest workflow since I don't think
anything is currently using those.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -2,6 +2,8 @@ name: Lint
 
 on: [push, pull_request]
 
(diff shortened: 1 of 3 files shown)


pjonsson:
I just tested introducing a type error and shuffling the imports around in this PR, and there were annotations shown in Github, so these permissions should be sufficient for what we currently use.

subnet001 (approved):

